PR #1, opened 2023-09-05
Rust go :rocket: :rocket: :rocket: 

tarkah:
A couple minor things while this repo gets setup

### Version 

We can use this built-in env var to get the package version: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates. If we want to include git hash, we can easily leverage `build.rs` to set an env var w/ that hash value. I can add this in a follow-up unit of work.

### Naming Convention

This one is a bit subjective, but I want to throw it out there. Rust modules offer a nice namespace so we shouldn't need to have redundant module names in the types or functions themselves:

```rs
mod list;

list::command()

// vs

use list::list_command; // redundant

mod list;

list_command()
```

### DIFF
--- a/src/cli/install.rs
+++ b/src/cli/install.rs
@@ -4,7 +4,7 @@
 
 use clap::Command;
 
-pub fn install_command() -> Command {
+pub fn command() -> Command {
     Command::new("install")
         .about("Install packages")
         .long_about("Install the requested software to the local system")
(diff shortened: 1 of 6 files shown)


ikeycode:
Solid stuff - thanks!

PR #8, opened 2023-09-09
TUI

tarkah:
Initial TUI! Still need to clean up some things and add our own progress bar implementation, but it's a pretty minimal async runtime and works pretty well.

https://github.com/serpent-os/moss-rs/assets/10239377/a9074591-e894-49b0-b50f-337c52098f7f


ikeycode:
Very nice ! What do we wanna do with the `examples` here ..?

tarkah:
> Very nice ! What do we wanna do with the `examples` here ..?

We can keep it or remove it now that we have `extract` hooked up.

I think this is in a good spot. I've only re-exported the ratatui primitives that are currently used. Ideally in the future we don't expose _any_ ratatui types but I can lock this down later.

ikeycode:
DO IT

PR #14, opened 2023-09-15
Add repository

tarkah:
Adds a new `repository` module w/ the ability to manage those via a `repository::Manager`. Creating a `Manager` will load all repo config files in admin `/etc` or vendor `/usr/share` paths . The manager allows you to add repositories (which creates an admin config file) and "refresh" the repositories index file to disk and meta db.

### TODO

- [x] Cleanup code
- [x] Fix db error when file doesn't exist
- [x] Truncate db tables before refresh (we can probably just recreate the file since sqlite :D) 
- [x] Get reqwest working w/ just `rustls`

image


moss/src/db/meta.rs
```diff
@@ -279,6 +276,8 @@ fn build_entry(package: package::Id, metadata: &[payload::Meta]) -> Result<Entry
     let hash = required_meta_string(metadata, payload::meta::Tag::PackageHash).ok();
     let download_size = required_meta_u64(metadata, payload::meta::Tag::PackageSize).ok();
 
+    let package = package::Id::from(hash.as_ref().unwrap_or(&name).clone());
+
```
tarkah, line 286: This works for remote / index files, but will this work for instances of using `meta_db`?

moss/src/remote/repository.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let mut out = File::create(out_path).await?;
+
+    let mut stream = resp.bytes_stream();
+
+    while let Some(chunk) = stream.next().await {
+        out.write_all(&chunk?).await?;
+    }
+
+    out.flush().await?;
```
tarkah: Streaming download / writing of index files :D

ikeycode (approved):
<insert appropriately intelligent sounding comment here>

PR #17, opened 2023-09-15
Refactor to package mod & simplify metadb API

tarkah:
Have metadb API simply work with `package::Metadata` type. Metadata module now encapuslates conversion logic between stone meta payload records and itself. Db
no longer needs to concern itself w/ any of this.

Consumer of meta db is now responsible over defining the `package::Id`. For our repository, the metadata must have the hash field populated.


moss/src/repository/manager.rs
```diff
@@ -184,16 +184,18 @@ fn update_meta_db(
...
         // We only care about meta payloads for index files
         let Ok(stone::read::Payload::Meta(payload)) = payload else {
-                continue;
-            };
+            continue;
+        };
+
+        let metadata = package::Metadata::from_stone_payload(&payload)?;
+        // Create id from hash of meta
+        let id = package::Id::from(metadata.hash.clone().ok_or(Error::MissingHash)?);
```
tarkah: Here's where we now construct the Id or error out if hash is missing. 

Now different consumers of meta-db can define their own requirements

ikeycode (approved):
v nice!!

PR #27, opened 2023-09-18
Add registry unit test for flags

tarkah:
Little unit test to confirm flag filtering works, i.e. `package.flags.contains(flag_parameter)`


ikeycode (approved):
Really glad you're adding these unit tests - will make upgrade logic far easier to integrate

PR #28, opened 2023-09-18
Add registry plugin

tarkah:


ikeycode (approved):
Only concern I have is potentially low memory constrained devices but chances are they won't be using list apis anyway, and more superimages.

tarkah:
> Only concern I have is potentially low memory constrained devices but chances are they won't be using list apis anyway, and more superimages.

We could eventually go full `Stream` API where each item is a separate async query. We'd just need to enforce package ordering within the plugin level (in sql statement), but that's not a huge deal.

PR #30, opened 2023-09-19
Feat/batch add

tarkah:
Batches payloads when inserting them into meta-db. This greatly improves the speed, while still providing a cap on memory usage up to the batch size.

Before / after:

image


ikeycode (approved):
We very much like this! In combination with the zstd preallocation we should be fighting fit =)

PR #31, opened 2023-09-19
Extract content in parallel

tarkah:
An example of doing `copy` in parallel from the stone content file to the store.

It appears to be around 20% faster based on some benchmarking


ikeycode (approved):
GIMME NOW

PR #46, opened 2023-09-27
Feat/meta db filter

tarkah:


ikeycode (approved):
Per our matrix chat - despite this potentially having some minimal overhead in single-provider scenarios - it locks down to a cleaner API and lays the groundwork for search APIs, etc.

It's also more memory efficient by returning less vectors

PR #54, opened 2023-10-11
Prevent layout dupes per package

tarkah:
Remove layout entries related to package id before batch adding to prevent dupes between installs


ikeycode (commented):
glad to see this rectified

ikeycode (approved):

PR #59, opened 2023-10-18
Handle file URL for request

tarkah:


ikeycode (approved):
This is beautiful for a reason you haven't elaborated on:

We now support local file:// repos and packages =)

PR #71, opened 2023-11-07
Mark selections as explicit

tarkah:
Breaking change, make sure to delete any previous `-D` root


ikeycode:
love it! upgrades here we come

ikeycode (approved):

PR #73, opened 2023-11-08
Hook up sync

tarkah:


ikeycode:
OK I'm adding context here for passerbys:

Rather than the traditional `upgrade` method employed by other package managers, we're opting to explicitly sync with the latest candidates as exposed by the cascading registry resolution.

Thus, I can add a repo, `sync`, get all the "updates". Remove the repo, or invert the priority, `sync` and go back to the old state. Unaccounted packages are unaffected and effectively pinned.

In future we'll also introduce pinned selections, which is basically mandatory to ensure sideloaded packages (via cobble) are residual.

ikeycode (approved):
Lets sync this!

PR #74, opened 2023-11-10
Feat/remote blit

tarkah:
`--to` on `sync` / `install` will output the blit to a custom target folder without recording any state. `sync` outputs the current state. `install` outputs _only_ the packages passed via `install`, which is nice for building ad-hoc container roots.


ikeycode:
Merging as required for boulder enabling (CAS!) - with note that duplicate handling is sorta "oops" right now, pending `--ignore-file-conflicts`

ikeycode (approved):

PR #77, opened 2023-11-15
Add stone writer

tarkah:


ikeycode (approved):
I had to wait for a merge window to get this in. In other projects thats a highly specific timeframe, in mine it was waking up before the kids need to go to school and type this reply as quietly as humanly possible for fear of awakening them before breakfast was prepared :3

#sendhelp

PR #82, opened 2023-11-21
Add workflow for nightly builds

der-eismann:
Creates releases with a binary of latest moss:
https://github.com/serpent-os/moss-rs/releases/tag/nightly-20231121-164115
Also only keeps the last 5 versions.


ermo (approved):
Ikey told me to merge, so approving. =)

PR #87, opened 2023-11-22
Feat/boulder

tarkah:


GZGavinZhao:
Would you mind adding support for the `conflicts` field in `package.yml` as well? Its specification should be exactly the same as the `rundeps` field (a list of providers) except that its Meta kind is Conflicts.

ikeycode:
As discussed on Matrix this needs merging **asafp** to ensure we have a baseline to operate against.
Fantastic work @tarkah as always. Let's get this road on the show.

PR #94, opened 2023-12-04
Initial support for conflicts

GZGavinZhao:
- dag: add neighbors_outgoing/incoming and scc
- moss: initial support for conflicts


crates/moss/src/db/meta/mod.rs
```diff
@@ -449,6 +477,26 @@ impl Database {
             .await?;
         }
 
+        // Conflicts
+        let conflicts = packages
+            .iter()
+            .flat_map(|(id, meta)| meta.conflicts.iter().map(move |conflict| (id, conflict)))
+            .collect::<Vec<_>>();
+        println!("conflicts: {conflicts:?}");
```
tarkah, line 485: Reminder to remove this `println`

crates/moss/src/package/meta.rs
```diff
@@ -4,10 +4,10 @@
 
 use std::{collections::HashSet, fmt};
 
-use stone::payload;
+use stone::payload::{self};
```
tarkah: Nit: can remove `::{self}`

crates/moss/src/package/meta.rs
```diff
@@ -254,6 +258,22 @@ fn meta_provider(meta: &payload::Meta) -> Option<Provider> {
     }
 }
 
+fn meta_conflict(meta: &payload::Meta) -> Option<Conflict> {
+    // Is there a way to idiomatically combine those in Rust?
+    if meta.tag == payload::meta::Tag::Conflicts {
+        if let payload::meta::Kind::Provider(kind, name) = meta.kind.clone() {
```
tarkah: You could do a `match`:

```rs
match (meta.tag, meta.kind) {
  (Tag::Conflicts, Kind::Provider(kind, name) => Some(Conflict { .. }),
  _ => None,
}
```

crates/moss/src/registry/transaction.rs
```diff
@@ -137,9 +139,116 @@ impl<'a> Transaction<'a> {
...
+            let mut visited: HashSet<package::Id> = HashSet::default();
+
+            for condition_index in component {
+                let (package_id, installed) = graph.get_node_from_index(condition_index);
+
+                if visited.contains(package_id) {
+                    let reasons: Vec<package::Id> = graph
+                        .neighbors_outgoing(&(package_id.clone(), false))
+                        .filter_map(|revdep| {
+                            if let (revdep_id, true) = revdep {
```
tarkah: Shouldn't this be neighbors incoming for the reverse dependency? Since we setup the connection from true to false and we're querying the false (conflict) node here?

crates/moss/src/registry/transaction.rs
```diff
@@ -137,9 +139,116 @@ impl<'a> Transaction<'a> {
...
+                    // Is this how I rethrow/rereturn an error in Rust?
+                    Err(e) => return Err(e),
+                }
+            }
+        }
+
+        let components = graph.scc();
+        let mut conflicts: Vec<(package::Id, Vec<package::Id>)> = vec![];
+        for component in components {
+            let mut visited: HashSet<package::Id> = HashSet::default();
```
tarkah: Are we creating any strongly connected components? Our DAG implementation prevents cycles so we shouldn't have any. With the additional key of true / false in the node, I can't see how they'd be created anyways. I suspect this will have a similar output to just a topological sort?

GZGavinZhao:
I found a problem with this algorithm, gonna fix it later this week.

GZGavinZhao:
@ikeycode @ermo @tarkah I want to at least make the `conflicts` field recognized/stored in `moss` before it's too late. We can work on how to best detect/report the conflicts in a separate PR. I'll try to rebase and split up this PR in the next few days.

GZGavinZhao:
See #248.

PR #97, opened 2023-12-12
Feat/boulder breakpoint

tarkah:
Adds support for `%break` & `%break_exit` actions in recipe scripts which dump the user into a chroot at that point of execution.

Example:

https://github.com/serpent-os/moss-rs/assets/10239377/4bf1d419-e23e-4eaa-922b-aec88f1c8e22

### DIFF
--- a/crates/stone_recipe/src/script.rs
+++ b/crates/stone_recipe/src/script.rs
@@ -21,11 +21,21 @@ use crate::{macros::Action, Macros};
 pub struct Parser {
     actions: HashMap<String, Action>,
     definitions: HashMap<String, String>,
+    env: Option<String>,
 }
 
 impl Parser {
     pub fn new() -> Self {
         Self::default()
+    }
+
+    /// Env is parsed and prependend to the beginning of each
+    /// [`Command::Content`]
+    pub fn env(self, env: impl ToString) -> Self {
+        Self {
+            env: Some(env.to_string()),
+            ..self
+        }
     }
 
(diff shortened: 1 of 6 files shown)


crates/stone_recipe/src/script.rs
```diff
@@ -126,7 +216,11 @@ fn tokens(input: &str, f: impl FnMut(Token) -> Result<(), Error>) -> Result<(),
     ));
 
     let token = alt((
-        map(action, Token::Action),
+        map(action, |action| match action {
+            "break" => Token::Break(Breakpoint { exit: false }),
```
ermo: Nit: Would it bother you a lot if I asked if this could be explicitly made `%break_continue` (or `%break_noexit`) to match the `%break_exit` action form (action -> consequence)?

This is _only_ about making it obvious from the action what its consequences will be. `%break` on its own is IMHO not sufficient to help the user understand what will happen if they use it?

ermo (changes requested):
Looks like it also needs a rebase.

Feels like a good PoC starting point for this feature, bang-for-the-buck wise to me. =)

crates/stone_recipe/src/script.rs
tarkah: 11950d2 agreed :+1:

tarkah:
> Looks like it also needs a rebase.

Rebased :+1:

ermo (approved):
This is awesome. LGTM!

PR #99, opened 2023-12-15
moss/inspect: detect dependencies/provides based on meta tag

GZGavinZhao:
Now that we have conflicts, detecting dependencies/provides solely based on the meta kind is not enough. We should use the meta tag to detect them.

Also it seems like the previous implementation got providers and dependencies in reverse? This also fixes that. Tested on the `boulder` stone to make sure that providers/dependencies/conflicts are correctly detected.

### DIFF
--- a/crates/moss/src/cli/inspect.rs
+++ b/crates/moss/src/cli/inspect.rs
@@ -47,9 +47,10 @@ async fn inspect(paths: Vec<PathBuf>) -> Result<(), Error> {
 
             let mut layouts = vec![];
 
-            // Grab deps/providers
+            // Grab deps/providers/conflicts
             let mut deps = vec![];
             let mut provs = vec![];
+            let mut cnfls = vec![];
 
             match payload {
                 PayloadKind::Layout(l) => layouts = l.body,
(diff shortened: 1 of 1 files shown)


crates/moss/src/cli/inspect.rs
```diff
@@ -92,6 +98,12 @@ async fn inspect(paths: Vec<PathBuf>) -> Result<(), Error> {
                     println!("    - {prov}");
                 }
             }
+            if !confs.is_empty() {
+                println!("\n{:width$} :", "Conflicts", width = COLUMN_WIDTH);
+                for conf in confs {
```
ermo: I'm not sure I'm keen on overloading "conf" and "confs" since those are usually used for to mean configuration and configurations respectively.

Would e.g. `cnfl` and `cnfls` work instead perhaps?

ermo (dismissed):

crates/moss/src/cli/inspect.rs
GZGavinZhao: That's the variable name I was looking for. Thanks!

ermo (commented):
LGTM!

ermo (approved):
LGTM!

PR #100, opened 2023-12-16
Feat/breakpoint linenumbers

tarkah:


crates/boulder/src/builder.rs
```diff
@@ -176,14 +179,24 @@ impl Builder {
...
+                                    )
+                                    .map(|line_num| format!(" at line {line_num}"))
+                                    .unwrap_or_default();
+
                                     println!(
-                                        "\n{} {}",
+                                        "\n{}{} {}",
                                         "Breakpoint".bold(),
+                                        line_num,
                                         if breakpoint.exit {
```
ermo, line 195: Suggestion: Make this more "principle of least astonishment"-y by switching the text to something like:

- (exit) -> (build will exit on shell exit)
- (continue) -> (build will continue on shell exit)

... perhaps?

ermo (commented):
Other than the nit, I think I like this.

crates/boulder/src/builder.rs
```diff
@@ -176,14 +179,24 @@ impl Builder {
...
                                             "(exit)".dim()
```
ermo, line 196: Suggestion: Make this more "principle of least astonishment"-y by switching the text to something like:

    (exit) -> (build will exit on shell exit)
    (continue) -> (build will continue on shell exit)

... perhaps?

ermo (commented):
Other than the nit, I like this (I might be repeating myself here!)

ikeycode (approved):

PR #108, opened 2023-12-19
Add boulder format

tarkah:
Applies our opinionated formatting (aligned & padded colons)


tarkah:
This doesn't preserve comments #fail, moving to draft until I think about this more.

PR #110, opened 2023-12-19
Add boulder update

tarkah:
image


tarkah:
Draft until we can figure out a better solution than deserializing since that strips non-data such as comments :/

ermo (approved):
I see no reason to _not_ approve this, as we need wider testing of it.

Issue #111, opened 2023-12-20
moss: cli panics when ran with --directory but no other arguments


it-a-me:
Running `moss -D /soroot` panics at 'crates/moss/src/cli/mod.rs:89:14' unreachable!.  It panics because --directory counts as a argument but not a subcommand

PR #112, opened 2023-12-20
print help when the only argument is --directory

it-a-me:
This should fix issue #111 where the cli panics with supplied with --directory but no subcommand.  If no subcommand is specified, moss now defaults to printing help rather than panics


tarkah (approved):
Thanks for the fix!

ikeycode (approved):

ikeycode:
thanks!

PR #113, opened 2023-12-21
container: Add README.md

livingsilver94:


tarkah:
I like this! It gives a high level overview without too much detail, a perfect primer to make sense of the underlying code on further review.

crates/container/README.md
```diff
@@ -0,0 +1,32 @@
+# container
+
+container is a crate to start a basic rootless container. Its goal is to create a safe and isolated environment where to build and/or test packages. Environment isolation is desired in that:
+
+  * We don't want to pollute the *host* system with temporary artifacts
+  * We want to work in a reproducible environment, for everyone (~~"it worked on my machine!"~~)
+  * Certain applications may run arbitrary code while building. We want to be able do to that securely.
+
+Emphasis is put on **rootless** container. Root permission must not be necessary, both for security and to reach a broader audience (including parent rootless containers, like Docker or Podman!).
```
tarkah, line 9: Just an FYI, `container` works when invoked as root as well. Boulder can be run as root and it runs this container w/out the addtl. rootless bits of logic (no id mapping necessary). 

It's not the "preferred" way, but it's still a supported way

tarkah (approved):

ikeycode (approved):
win!

PR #129, opened 2024-01-25
Initial inclusion of pre-activation triggers

ikeycode:
Thoughts and suggestions welcome, this lands the initial `pre-triggers` as part of a new unified `postblit` process that will eventually grow to absorb tasks like os-release management.

We currently have no caching mechanism, so each trigger is run fresh for each new transaction. Some scaffolding is
put in place to know which paths are involved in the fnmatch-based glob, so that in future we can capture the input hashes (CA xxhash) to "save" trigger executions.

Note this PR yields a writable `/usr` and `/etc`, however the plan is to make `/etc` read-only. This is currently broken and needs investigation once we have a baseline,


ikeycode:
For testing this PR note that `shared-mime-info`, `glib2`, `gdk-pixbuf`, all now contain triggers in `volatile`.

ikeycode:
yolo

crates/moss/Cargo.toml
```diff
@@ -32,3 +32,5 @@ tokio-util.workspace = true
 thiserror.workspace = true 
 url.workspace = true
 xxhash-rust.workspace = true
+triggers = { version = "0.1.0", path = "../triggers" }
+container = { version = "0.1.0", path = "../container" }
```
tarkah, line 36: Nit: we have local crates grouped above

crates/moss/src/client/postblit.rs
```diff
@@ -0,0 +1,127 @@
...
+#[derive(Deserialize, Debug)]
+struct PreTrigger(Trigger);
+
+impl config::Config for PreTrigger {
+    fn domain() -> String {
+        "pre".into()
+    }
+
+    fn merge(self, other: Self) -> Self {
+        todo!()
```
tarkah, line 37: 👀

crates/triggers/src/lib.rs
```diff
@@ -26,40 +38,116 @@ pub enum Error {
...
+                self.hits
+                    .iter()
+                    .filter(move |(id, _, _)| i == *id)
+                    .collect::<Vec<_>>()
+            })
+            .flatten()
+            .collect::<Vec<_>>();
+
+        let mut runnables: BTreeMap<format::Handler, TriggerCommand> = BTreeMap::new();
+        for (_, hit, handler) in built_triggers {
```
tarkah, line 135: Do we no longer care about dependency order when adding to runnables? format::Handler seems to be ordered basically off command name

crates/moss/src/client/postblit.rs
ikeycode: never used as we load all :D

PR #132, opened 2024-02-05
Triggers cleanup

tarkah:
A few small  things I noticed when reviewing triggers.

`Config::merge` doesn't make sense at the API level as it causes `unimplemented`. Default API should just always return an array of configs and merging can be handled by the consumer.


ikeycode (approved):
merci - much appreciated =)

PR #134, opened 2024-02-09
moss: Add a CACHEDIR.TAG to `.moss/cache`

hunger:
Just a small patch adding in a CACHEDIR.TAG into `.moss/cache` to prevent modern backup tools from backing up data in that directory.

### DIFF
--- a/crates/moss/src/installation.rs
+++ b/crates/moss/src/installation.rs
@@ -188,4 +188,17 @@ fn ensure_dirs_exist(root: &Path) {
     ] {
         let _ = fs::create_dir_all(path);
     }
+    ensure_cachedir_tag(&moss.join("cache"));
+}
+
+fn ensure_cachedir_tag(path: &Path) {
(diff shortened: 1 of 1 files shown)


ikeycode:
Seems legit :)

ikeycode (approved):

crates/moss/src/installation.rs
```diff
@@ -188,4 +188,17 @@ fn ensure_dirs_exist(root: &Path) {
     ] {
         let _ = fs::create_dir_all(path);
     }
+    ensure_cachedir_tag(&moss.join("cache"));
+}
+
+fn ensure_cachedir_tag(path: &Path) {
+    let cachedir_tag = path.join("CACHEDIR.TAG");
+    if !cachedir_tag.exists() {
+        let _ = std::fs::write(
```
tarkah, line 197: Nit: `fs` module is already imported, so you can drop `std::`

PR #138, opened 2024-02-13
List orphan removal from sync

tarkah:
And allow sync to run if no upgrades exist, but removals do.

This doesn't change the behavior of sync, as it previously removed orphans. We just didn't present that information to the user nor allow it to run if only removals exists & not upgrades


ikeycode (approved):
way more useful!

PR #139, opened 2024-02-13
Don't unpack dupe file

tarkah:
I'm not actually sure this fixes anything as it's hard to reproduce the Text file busy IO error in VM


ikeycode:
To the outside observer: We had a follow up conversation where we decided mere file existence wasn't enough, and decided that a hash claim store would be the best bet, so this PR was modified to suit.

ikeycode (approved):

PR #140, opened 2024-02-14
Print installed files for moss info -f

tarkah:
Example of difference between regular file vs symlink

image

### DIFF
--- a/crates/moss/src/client/mod.rs
+++ b/crates/moss/src/client/mod.rs
@@ -468,6 +468,26 @@ impl Client {
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


crates/moss/src/client/mod.rs
```diff
@@ -468,6 +468,26 @@ impl Client {
         Ok(())
     }
 
+    /// Build a [`vfs::Tree`] for the provided packages
```
ikeycode, line 471: very nice use of vfs =)

ikeycode (approved):

PR #146, opened 2024-02-26
[NFC] Move apps to the top level directory

livingsilver94:
I think there should be a separation between apps and libraries, especially for newcomers.


ermo:
On the face of it, seems a pretty sensible suggestion...

Cargo.toml
```diff
@@ -24,11 +20,23 @@ futures = "0.3.30"
...
+    "term",
+] }
 once_cell = "1.19.0"
 petgraph = "0.6.4"
 rayon = "1.8"
-reqwest = { version = "0.11.23", default-features = false, features = ["rustls-tls", "stream"] }
+reqwest = { version = "0.11.23", default-features = false, features = [
+    "rustls-tls",
+    "stream",
+] }
```
tarkah: Nit: personally not a fan of wrapping on objects here. I'd rather see these keys under [dependencies.reqwest] table below. Otherwise I think these should be oneliners.

I'd personally just disable toml formatting

tarkah:
Agree with the structure change, it promotes our binary crates. I just have a nit about toml formatting

Cargo.toml
livingsilver94: Done. I personally don't like what I did because the configuration of those dependencies is now visually separated from the `[workspace.dependencies]` block, but here you are.

ikeycode (approved):

PR #147, opened 2024-02-27
Task/simplify db

tarkah:
Removes a lot of boilerplate code in preference to derivable From, Into, Display implementations via `derive_more` and `strum` crates.


ikeycode (approved):

ikeycode:
nice simplification =)

PR #148, opened 2024-02-28
Drop the bitflags dependency

livingsilver94:
This PR replaces the bitflags functionality with an in-house implementation.

I think we all agree on reducing the number of dependencies where it makes sense, and I also think bitflags is one dependency worth removing. It is based on a `macro_rule` which is able to escape the original Rust syntax (think of that `: u8` syntax to specify that flags are bytes. That's a C++ notation!), making code a little confusing. `macro_rules` have too much freedom of toying with the code, IMO.

The new `Flags` struct is a type-safe and value-safe collection of flags. It is not possible to build an invalid collection of Flags. The idea is based on `std::fs::OpenOptions`.

Did you note that `bits()` private method? That's a micro-optimization. While I could implement the `contains()` method using pure boolean logic, with Compiler Explorer I noticed that that would introduce conditional jumps in the assembly. For the sake of the CPU pipeline, I chose to convert the single flags back into an integer for the comparison, thus avoiding jumps altogether.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1293,7 +1293,6 @@ dependencies = [
 name = "moss"
 version = "0.1.0"
 dependencies = [
- "bitflags 2.4.2",
  "bytes",
  "chrono",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ resolver = "2"
 edition = "2021"
 
 [workspace.dependencies]
-bitflags = "2.4.1"
 bytes = "1.5.0"
 chrono = "0.4.30"
 clap = { version = "4.4.11", features = ["derive", "string"] }
--- a/moss/Cargo.toml
+++ b/moss/Cargo.toml
@@ -10,7 +10,6 @@ stone = { path = "../crates/stone" }
 tui = { path = "../crates/tui" }
 vfs = { path = "../crates/vfs" }
 
-bitflags.workspace = true
 bytes.workspace = true
 chrono.workspace = true
 clap.workspace = true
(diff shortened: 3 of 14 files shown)


crates/moss/src/package/mod.rs
```diff
@@ -67,20 +66,59 @@ impl Ord for Package {
...
+impl Flags {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns a copy of [`Flags`] with available set to true.
+    pub fn with_available(&self) -> Self {
+        Self {
+            available: true,
+            ..*self
```
tarkah: Builder pattern is typically either `&mut self` or `self`. I think for our usage `self` makes the most sense, and you'll eliminate the `* / Copy`

tarkah (changes requested):
Awesome! For me the win is not having LSP breakdown when doing go-to-definition on `contains` or any of the other implementations due to macro usage. Plus it's easier to reason about in general, I like. 

One small comment around builder pattern.

crates/moss/src/package/mod.rs
livingsilver94: Makes sense! That means that `self` is a copy, instead of a consumption of the object?

livingsilver94: Because Self implements Copy, I mean.

tarkah: Yeah apologies, we're still copying here

livingsilver94:
Oops, made a mess with git history. Fixing.

tarkah (approved):

tarkah (approved):

PR #150, opened 2024-02-28
boulder: Bring data/ up to parity with boulder-d

ermo:
Note that a new x86_64-v3x.yml target definition was added in preparation for the point where we will begin building the scaffolding to use it.

### DIFF
--- a/boulder/data/macros/actions/cmake.yml
+++ b/boulder/data/macros/actions/cmake.yml
@@ -28,6 +28,14 @@ actions:
         dependencies:
             - ninja
 
+    # Run testsuite with ctest
+    - cmake_test:
+        command: |
+            ninja test -v -j "%(jobs)" -C "%(builddir)"
+        dependencies:
+            - cmake
+            - ninja
+
 definitions:
 
     # Default cmake options as passed to cmake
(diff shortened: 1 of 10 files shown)


boulder/data/default-x86_64.conf
```diff
@@ -0,0 +1,20 @@
+- default-x86_64:
+    collections:
```
tarkah: We use `yaml` file extension for our config files now since there... yaml :)

tarkah: Also, we don't use the weird array / map nesting, so the `-` should be dropped from the map keys

tarkah: Shit.. we already have `boulder/data/profile.d/default-x86_64.yaml` :)

ermo: the `.yml` vs. `.yaml` thing is getting on me nerves (I personally prefer `.yaml`).

livingsilver94: The standard technically imposes `.yaml`. We should just follow that extension and call it a day, it's a no-brainer.

ermo: ... We should probably use `.yaml` project wide as well then (assuming this doesn't make @ikeycode blow a gasket!)

tarkah (approved):

PR #156, opened 2024-03-01
crates/tui: Simple refactor

livingsilver94:
This is mostly code style you may not agree with, but I think the replacement of the reader and writer wrappers is a win.

- `TermSize` now implements the Default trait, and the previous `get_size()` function is now `TermSize`'s constructor.
- `ProgressReader` and `ProgressWriter` are replaced by indicatif's `wrap_read()` and `wrap_write()`:
  - Shorter code
  - Prevents a previously possible division by zero
  - Those wrapper expose more methods than `read()` and `write()` for convenience

### DIFF
--- a/crates/tui/src/lib.rs
+++ b/crates/tui/src/lib.rs
@@ -2,108 +2,38 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use std::io::{Read, Write};
-
-pub use self::reexport::*;
 pub use self::styled::Styled;
+pub use dialoguer;
+pub use indicatif::*;
 
 pub mod pretty;
 mod styled;
 
-const DEFAULT_TERM_SIZE: (u16, u16) = (80, 24);
-
-/// Simple terminal constraints wrapping
+/// The size of a terminal emulator window.
 pub struct TermSize {
+    /// Width (or number of columns) of the terminal, in characters.
     pub width: usize,
+    /// Height (or number of rows) of the terminal, in characters.
     pub height: usize,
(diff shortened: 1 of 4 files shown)


crates/tui/src/lib.rs
```diff
@@ -2,107 +2,36 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use std::io::{Read, Write};
-
-pub use self::reexport::*;
```
livingsilver94, line 7: I removed this mod entirely, but should you choose to keep it, Rust usually calls it `prelude`. See `std::io::prelude` for example.

tarkah: `reexport` isn't an exposed module, we're exposing the _contents_ of `reexports` root exports of this crate. It was just a trivial way to define all the things we wanted re-exported from other crates in one place, and have the all available from the _root_ of the crate

crates/tui/src/lib.rs
```diff
@@ -2,107 +2,36 @@
...
-
 pub mod pretty;
+pub use crossterm::style::Stylize;
+pub use dialoguer;
+pub use indicatif::*;
```
tarkah: Nit: We put all import lines before declaring `mod` across all the crates. For consistency, these should go above. Also a space in-between imports / modules

tarkah (approved):
Nice catch on `wrap_read / write`! That cleans up a bunch :+1:  Just one small nit

tarkah (approved):

PR #164, opened 2024-03-05
crates/triggers: Add OsEnv

livingsilver94:
Nothing really useful here until we make use of this OsEnv, but meanwhile we wait for the triggers crate to be completed, I feel like I should get my feet wet.


livingsilver94:
Closing. I'm gonna merge this edit into fnmatch-filesystem.

PR #167, opened 2024-03-06
Feature/state management

tarkah:
Adds `moss state remove / active / activate` and fixes `prune` to only consider before the current active state


ermo (approved):
Tested with img-tests/desktop/img.sh script.

Worked like a charm!

PR #170, opened 2024-03-07
Example of eyre use for extract command

tarkah:
https://docs.rs/color-eyre/latest/color_eyre/

### DIFF
--- a/moss/src/installation.rs
+++ b/moss/src/installation.rs
@@ -224,8 +224,8 @@ fn ensure_cachedir_tag(path: &Path) {
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("Root is invalid")]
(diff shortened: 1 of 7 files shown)


moss/src/installation.rs
```diff
@@ -224,8 +224,8 @@ fn ensure_cachedir_tag(path: &Path) {
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("Root is invalid")]
+    #[error("root is invalid")]
     RootInvalid,
-    #[error("Cache dir is invalid")]
+    #[error("cache directory is invalid")]
     CacheInvalid,
```
tarkah, line 230: Should our root library errors start capitalized or lowercase? :thinking: And then should all context be lowercase? Consistency here lends towards a nicer UX

PR #176, opened 2024-03-09
Fix/empty content binary stone

tarkah:


ikeycode:
I got some questions on this. What happens if the patterns yield an empty package, or we want an empty metapackage.. or .. ? :)

tarkah:
Should "just work" but I've only tested with a package that only produced a symlink (no content, but 1 layout record).

We should test truly meta only package from boulder -> moss to ensure no layout entries doesn't also break (which it might)

ermo:
@ikeycode If you could review this ASAP, that'd be super useful. :sweat_smile:

ikeycode (approved):

PR #188, opened 2024-03-18
boulder: Add -b flag to specify build release number

tarkah:


ermo (approved):
Keeping this simple for now (no checks beyond it being a valid NonZeroU64) IMHO makes sense. Then we can leave the checking to a separate set of layers.

Issue #34, opened 2023-09-20
Implement `moss search (sr)`


ikeycode:
Per D version, support searching of the Registry by name, summary, description.
Initially we can just use the deserialised values, but the most optimal approach will def. be plugging up some decent sqlx queries and operating on a `SearchMode` enum

livingsilver94:
- Do we want to look for keywords in names, summaries, descriptions, and only those?
- Is the search case insensitive? Or do we want an option for that?
- Do we want to rely on the sqlite db for the search?

ermo:
> * Do we want to look for keywords in names, summaries, descriptions, and only those?
> 

Yes, I believe that we do.

Compared to eopkg, it would be nice if the search automagically listed the part of the field(s) that match the search?

>     * Is the search case insensitive? Or do we want an option for that?
> 

Realistically, it's smart to make it case insensitive by default, then have an option to care about case sensitivity,  effectively narrowing down the search interactively?

>     * Do we want to rely on the sqlite db for the search?

Probably?

livingsilver94:
I edited the top level comment with the plan we agreed upon on the Matrix chat.

livingsilver94:
Can this be closed? Is it only partially resolved? In that case, we could create subissues.

PR #200, opened 2024-04-09
moss: Allow to search for packages with one keyword

livingsilver94:
First implementation step of #34.

Outputs:
```text
./target/debug/moss -D ~/Personal/SerpentOS/SOSROOT/ search gcc  -i
gcc       GNU C++ standard library and compiler 
```
```text
fforni@fedora:~/Personal/SerpentOS/moss$ ./target/debug/moss -D ~/Personal/SerpentOS/SOSROOT/ search gcc  
gcc-32bit-devel       Provides development files for gcc-32bit gcc-dbginfo                          Debugging symbols for gcc 
gcc-32bit            Provides 32-bit runtime libraries for gcc gcc                      GNU C++ standard library and compiler 
gcc-32bit-dbginfo              Debugging symbols for gcc-32bit gcc-devel                            Development files for gcc 
```

(Github destroyed the latter output. Inside a terminal it's well aligned). Also, package names are **bold**.

I'm not sure we should print 2 columns here. `moss search` output is supposed to be greppable I think.

### DIFF
--- a/moss/src/cli/search.rs
+++ b/moss/src/cli/search.rs
@@ -1,0 +1,90 @@
+// SPDX-FileCopyrightText: Copyright © 2020-2024 Serpent OS Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
+use clap::builder::NonEmptyStringValueParser;
+use clap::{Arg, ArgMatches, Command};
+
+use moss::client;
+use moss::package::{self, Name};
+use moss::{environment, Client, Installation};
+use tui::pretty::{print_columns, ColumnDisplay};
+use tui::Styled;
+
+const ARG_KEYWORD: &str = "KEYWORD";
+const FLAG_INSTALLED: &str = "installed";
+
+/// Returns the Clap struct for this command.
+pub fn command() -> Command {
+    Command::new("search")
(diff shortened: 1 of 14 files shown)


moss/src/cli/search.rs
```diff
@@ -0,0 +1,90 @@
...
+    name: Name,
+    summary: String,
+}
+
+impl ColumnDisplay for Output {
+    fn get_display_width(&self) -> usize {
+        // TODO: calculate the number of graphemes, not bytes.
+        // Now we are assuming name and summary are ASCII.
+        self.name.to_string().len() + self.summary.len() + COLUMN_SPACING
+    }
```
tarkah, line 79: Agreed we shouldn't show multiple package columns in this output. So this is either the wrong API to use or needs to be tweaked to add a max_num_columns it can use internally and set to one here

tarkah (commented):
This looks great! Let's fix the multi column output and I think this is good to go for phase 1.

tarkah (approved):
LGTM! Nice work

ikeycode (approved):
lgtm! clippy error just seems to be a rebase issue

PR #209, opened 2024-04-21
boulder: include buildrel in jsonc manifest

ermo:
This makes it easier to see when `boulder build -bN` has been used to build a set of artefacts.

Example:

```diff
diff --git a/x/xorriso/manifest.x86_64.jsonc b/x/xorriso/manifest.x86_64.jsonc
index e535ffc3..baef6688 100644
--- a/x/xorriso/manifest.x86_64.jsonc
+++ b/x/xorriso/manifest.x86_64.jsonc
@@ -38,5 +38,6 @@
        },
        "source-name": "xorriso",
        "source-release": "1",
-       "source-version": "1.5.6.pl02"
+       "source-version": "1.5.6.pl02",
+       "build-release": 2
 }

```

### DIFF
--- a/boulder/src/package/emit/manifest/json.rs
+++ b/boulder/src/package/emit/manifest/json.rs
@@ -6,6 +6,7 @@ use std::{
     collections::{BTreeMap, BTreeSet},
     fs::File,
     io::Write,
+    num::NonZeroU64,
     path::Path,
 };
 
@@ -18,6 +19,7 @@ use crate::{package::emit, Recipe};
 pub fn write(
     path: &Path,
     recipe: &Recipe,
+    build_release: NonZeroU64,
     packages: &BTreeSet<&emit::Package>,
     build_deps: &BTreeSet<String>,
 ) -> Result<(), Error> {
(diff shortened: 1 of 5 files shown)


boulder/src/package/emit/manifest/json.rs
```diff
@@ -64,6 +75,7 @@ pub fn write(
         source_name: recipe.parsed.source.name.clone(),
         source_release: recipe.parsed.source.release.to_string(),
         source_version: recipe.parsed.source.version.clone(),
+        build_release: metadata_build_release.to_string(),
```
tarkah: Do we need this to serialize as a string for any reason?

tarkah (approved):
LGTM

boulder/src/package/emit/manifest/json.rs
ermo: Erhm. I think I added that because `source_release` also does it?

tarkah: Might as well go u64 since that's what it is. It's never going to be stringy

tarkah (approved):

PR #218, opened 2024-04-24
Feat/cobble

### LINKED ISSUE #205: moss: Enable cobble-plugin so operations work on .stones directly
> **ermo**
> `moss info linux-kvm-6.8.6-66-1-x86_64.stone` <- what is the correct invocation for moss info to work for that?
> or will only `moss inspect` ever work for that purpose?
> because if `moss info` is not supposed to work on arguments ending in `.stone`, it might be nice to have moss suggest to use `moss inspect` or simply doing that instead?
> Also, I would like to be able to do a `moss extract some.stone` to a folder, without having to create a `.moss/` dir. Is this equivalent to having an alias for `moss install --to`?
> use case: I want to inspect the initrd from the kernel .stone I just built, before I push it to prod. I don't want to do a full `.moss/` root for this initially.
> `$ moss install --to /tmp/linux-kvm/ linux-kvm-6.8.6-66-1-x86_64.stone`
> `Error: install: client: db: diesel connection: Unable to open the database file`
> oh fuck off and just do it, will you moss?!
> (sorry for the outburst, was just trying to do my due diligence before pushing)
> **Ikey Doherty [BST/UTC+1]**
> So tldr we need cobble working 
> And for local arguments only prevent initialisation of system plugins
> Right now you need to use a local repo as I never bothered to finish porting cobble 
> It's not actually "hard" to do tbh
> **ermo**
> expected difficulty/annoyance level?
> Because if it's not hard, I'd love to have the ability to do the above.
> **Ikey Doherty [BST/UTC+1]**
> Basically we need access to the cobble plugin api via client and "preload" the stones into memory 
> Basically a metadb in ram
> Then we convert the input .stone paths into their package::Id
> Tldr we take our input args, filter out which end with .stone and exist 
> If remaining args is non empty, initialise other system plugins too
> **ermo**
> and this applies to which operations?
> like, would it also work for info?
> **Ikey Doherty [BST/UTC+1]**
> For install op, always init all plugins 
> Then from all paths map the vec to package::Id via client.sideload()
>> **ermo**
>> like, would it also work for info?
> Ya
> Tbh info should only init system plugins (non cobble) when it has args that aren't valid .stone paths 
> Then apply that logic to install, and extend to map those paths to package::Id and inject into normal install workflow
> Minimally disruptive

tarkah:
Resolves #205

Currently cobble installed packages are "ephemeral" and will get dropped on the next `sync`. Users should add the stones to a local repo to get persistence, or we need to come up with some idea of pinning cobbled stones.

### DIFF
--- a/moss/src/cli/info.rs
+++ b/moss/src/cli/info.rs
@@ -2,8 +2,10 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+use std::path::PathBuf;
+
 use clap::{arg, ArgMatches, Command};
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use moss::{
     client::{self, Client},
     environment,
@@ -35,26 +37,48 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         .collect::<Vec<_>>();
     let show_files = args.get_flag("files");
 
-    let client = Client::new(environment::NAME, installation)?;
+    let mut client = Client::new(environment::NAME, installation)?;
 
-    for pkg in pkgs {
-        let lookup = Provider::from_name(&pkg).unwrap();
+    // Partition between names and stone paths
+    let (pkg_names, local_stones): (Vec<String>, Vec<PathBuf>) = pkgs.into_iter().partition_map(|name| {
+        if name.ends_with(".stone") {
+            Either::Right(name.into())
+        } else {
+            Either::Left(name)
+        }
+    });
+
+    let mut candidates = vec![];
+
+    // Cobble local stones and add their packages
+    if !local_stones.is_empty() {
+        candidates.extend(client.cobble(&local_stones)?);
+    }
+
+    // Resolve & add all package names
+    for pkg in pkg_names {
+        let lookup = Provider::from_name(&pkg).map_err(|_| Error::ParseProvider(pkg.to_string()))?;
         let resolved = client
             .registry
             .by_provider(&lookup, Flags::default())
             .collect::<Vec<_>>();
+
         if resolved.is_empty() {
             return Err(Error::NotFound(pkg));
         }
-        for candidate in resolved {
-            print_package(&candidate);
 
-            if candidate.flags.installed && show_files {
-                let vfs = client.vfs([&candidate.id])?;
-                print_files(vfs);
-            }
-            println!();
+        candidates.extend(resolved);
+    }
+
+    // Print each candidate
+    for candidate in candidates {
+        print_package(&candidate);
+
+        if candidate.flags.installed && show_files {
+            let vfs = client.vfs([&candidate.id])?;
+            print_files(vfs);
         }
+        println!();
     }
 
     Ok(())
@@ -133,6 +157,8 @@ fn print_files(vfs: vfs::Tree<client::PendingFile>) {
 
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("Not a valid provider name: {0}")]
+    ParseProvider(String),
     #[error("No such package {0}")]
     NotFound(String),
     #[error("client")]
--- a/moss/src/cli/sync.rs
+++ b/moss/src/cli/sync.rs
@@ -188,7 +188,7 @@ fn resolve_with_sync(
             Resolution::Explicit => p.flags.explicit,
             Resolution::All => true,
         })
-        .map(|p| {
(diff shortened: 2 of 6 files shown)


ermo:
My test case (which doesn't work):

        Compiling diesel_migrations v2.1.0
        Finished packaging [optimized + debuginfo] target(s) in 3m 19s
    ermo@solbox:~/repos/serpent-os/moss [feat/cobble* +4 ~0 -0 !]
    $ target/packaging/moss -v
    moss 0.1.0 (e0b9035)
    ermo@solbox:~/repos/serpent-os/moss [feat/cobble* +4 ~0 -0 !]
    $ git log -1
    commit e0b903524734d649056b28e18802129b4ef3e630 (HEAD -> feat/cobble, origin/feat/cobble)
    Author: Cory Forsstrom <cforsstrom18@gmail.com>
    Date:   Wed Apr 24 13:34:47 2024 -0700
    
        moss: info: Show info for cobbled stones
    ermo@solbox:~/repos/serpent-os/moss [feat/cobble* +4 ~0 -0 !]
    $ sudo cp -v target/packaging/moss /usr/bin/
    'target/packaging/moss' -> '/usr/bin/moss'
    ermo@solbox:~/repos/serpent-os/moss [feat/cobble* +4 ~0 -0 !]
    $ moss -v
    moss 0.1.0 (e0b9035)
    (...)
    $ moss info 
    font-cantarell-0.303.1-1-2-x86_64.stone  manifest.x86_64.jsonc                    stone.yaml
    manifest.x86_64.bin                      monitoring.yml                           
    ermo@solbox:~/repos/serpent-os/recipes/f/font-cantarell [autobuild-rebuilds* +2 ~0 -0 !]
    $ moss info font-cantarell-0.303.1-1-2-x86_64.stone 
    Error: info: client: db: diesel connection: Unable to open the database file
    ermo@solbox:~/repos/serpent-os/recipes/f/font-cantarell [autobuild-rebuilds* +2 ~0 -0 !]
    $ moss -v
    moss 0.1.0 (e0b9035)

ikeycode:
Yeah the dropping of stones seems wrong to me tbh. Really sideloaded means "i'm pushing this in and above the defaults" so would need the concept of pinning by way of state selections I would imagine. Until the user unpins or restores the cobbled package (ie. "my system is cobbled together from these stones ...") we continue trying to resolve transactions until they inevitably break our dependency graph.

ikeycode:
A potential "workaround" here actually is layering cobble atop a repo plugin. Such that, any time we "sideload" a stone (or set of stones) we promote from their evaluation site into a staged cobble repository, which just so happens to have a higher priority than everything else.

Eliminates the need for pinning support, and deviates from the `apt` model by having users pin to a distribution *revision* or rollback instead.

ermo:
The only time cobble has interest for me is for when I'm doing quick tests for packaging validation purposes. Hence, to me, cobble should be explicitly designed to support that use-case.

That said, is there a use-case here we might use cobble to drive bootstrap logic via building "dirty" .stones outside of containment (so using local environemnt deps instead of boulder-controlled .stone deps)? If so, those are the two use-cases I could see as being useful for cobble.

To me, sideload equates creating a convenient shorthand using a local repo and managing that. That might imply that we're better off having `sideload` actually be a separate "group" of commands (or an alias...?) that copies .stones to the "sideload" (highest priority) local repo? That way, it does what it should, it works with the normal repo plugins _and_ it's controlled by "normal" moss logic re. sync etc.?

I think we need to be very careful re. using the cobble plugin only for very well defined edge cases, such that normal uses don't get caught in a trap they can't easily get out of if they use it.

ermo:
The other option here is that cobble completely removes the need for local repos to be indexed -- cobble takes care of that? In this scenario, cobble actually becomes _the_ mechanism for handling local repos, with the caveat that you can only have one "saved state" of a cobble repo and that's always "what it looks like now".

ermo:
> Eliminates the need for pinning support, and deviates from the `apt` model by having users pin to a distribution _revision_ or rollback instead.

I think we need to accept that with the idea of a versioned recipe being brought into being as a versioned moss-format repo version, we MUST pin to a distribution _revision_.

This also has implications for the `system-model` idea; it would consist of a revision (which could support a special version of `ANY`, which means "rolling") along with the tags/groups/packages that were valid for that revision. As soon as moss cannot sync to the specified model, it would show the tags/groups/packages that are no longer valid and ask for user input.

PR #219, opened 2024-04-26
moss: Add verify command

tarkah:
Resolves https://github.com/serpent-os/moss/issues/126

### TODO
- [x] Cleanup CLI output

### DIFF
--- a/moss/src/cli/state.rs
+++ b/moss/src/cli/state.rs
@@ -45,6 +45,7 @@ pub fn command() -> Command {
                     .value_parser(clap::value_parser!(u64)),
             ),
         )
+        .subcommand(Command::new("verify").about("Verify TODO"))
 }
 
 pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error> {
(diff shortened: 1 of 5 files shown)


moss/src/cli/state.rs
```diff
@@ -45,6 +45,11 @@ pub fn command() -> Command {
                     .value_parser(clap::value_parser!(u64)),
             ),
         )
+        .subcommand(
+            Command::new("verify")
+                .about("Verify TODO")
+                .arg(arg!(--verbose "Vebose output").action(ArgAction::SetTrue)),
+        )
```
tarkah, line 52: We should probably add this as a global / root CLI option w/ short as `v` and change `version` short to `V`.

@ermo thoughts?

ermo: @tarkah Yes! I was thinking the same thing last night actually.

ermo:
@tarkah Needs a rebase ^^'

ermo (approved):
I'd rather have this in than not so we can begin testing it. If it has issues, we can fix/tweak later.

PR #221, opened 2024-04-29
boulder: Always run x86 build before x64 one

ReillyBrogan:
As mentioned in the packaging channel, we should always run the x86 build before the x64 one so that the x64 install phase clobbers the x86 one and not vice-versa.


tarkah (approved):
LGTM thanks!

PR #224, opened 2024-04-29
boulder: Fix interpreter detection

ReillyBrogan:
This wasn't correct and prevented the glibc build from having a provides on the interpreters.


ermo:
@ikeycode This is "unbreak now" priority. Please check that it does what it should and then merge ASAP?

ermo (approved):
Turns out that legacy DLang boulder also [uses /usr/lib](https://github.com/serpent-os/libmoss/blob/19e0af8f84cd34c17adfbc24a2f46ff8463ffaad/source/moss/deps/analysis/elves.d#L163).

Approving this so we can move on.

PR #235, opened 2024-05-06
all: Switch to deterministic BTreeMap/BTreeSet

ReillyBrogan:
While testing ccache I noticed that cache hit rate was extremely low. After some debugging I noticed that the CFLAGS and other environmental variables and macro scripts were not being set deterministically. This is problematic for ccache as it hashes the compiler arguments as part of the key and the order matters to get the same key on a future run.

The cause of this non-determinism is the Rust stdlib HashMap and HashSet, for which the default hasher used a random seed. To solve this switch to BTreeMap and BTreeSet from the stdlib. Since non-determinism is just generally a bad idea for build tools and package management software replace ALL uses of HashMap and HashSet repo-wide.

Tested by building a package with ccache enabled and using `ccache --show-stats` in the package recipe to verify that the hit rate was now 100% and that the build stage time had dropped to just the time it spent linking.

### DIFF
--- a/boulder/src/architecture.rs
+++ b/boulder/src/architecture.rs
@@ -37,7 +37,7 @@ impl Architecture {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Display)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Display)]
 pub enum BuildTarget {
     #[display(fmt = "{_0}")]
     Native(Architecture),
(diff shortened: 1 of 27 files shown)


ermo:
@tarkah Your input on this would be appreciated. I don't see any reason why it _shouldn't_ go in, but maybe I'm missing something...?

tarkah:
This sounds good in theory. Let me dig through each change tomorrow to confirm there's no sneaky regressions due to change from Hash/Eq to Ord

ermo:
> This sounds good in theory. Let me dig through each change tomorrow to confirm there's no sneaky regressions due to change from Hash/Eq to Ord

Any luck with this... @tarkah ?

ikeycode (approved):

PR #238, opened 2024-05-09
boulder: Add SOURCE_DATE_EPOCH

ReillyBrogan:
Wire up SOURCE_DATE_EPOCH. This environmental variable will always be set in the buildroot and will be set according to the following priority:
- Check to see if the environmental variable is already set, if so propagate it
- Check the return value of `git log -1 --format="%at"`, which returns the timestamp of the last commit of the git repo
- Use the current time

### DIFF
--- a/boulder/src/recipe.rs
+++ b/boulder/src/recipe.rs
@@ -3,10 +3,12 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use std::{
-    fs, io,
+    env, fs, io,
     path::{Path, PathBuf},
+    process::Command,
 };
 
+use chrono::{DateTime, Utc};
 use thiserror::Error;
 
 use crate::architecture::{self, BuildTarget};
@@ -18,6 +20,7 @@ pub struct Recipe {
     pub path: PathBuf,
     pub source: String,
     pub parsed: Parsed,
+    pub build_time: DateTime<Utc>,
 }
 
 impl Recipe {
@@ -25,8 +28,14 @@ impl Recipe {
         let path = resolve_path(path)?;
         let source = fs::read_to_string(&path)?;
         let parsed = stone_recipe::from_str(&source)?;
+        let build_time = resolve_build_time(&path);
 
-        Ok(Self { path, source, parsed })
+        Ok(Self {
+            path,
+            source,
+            parsed,
+            build_time,
+        })
     }
 
     pub fn build_targets(&self) -> Vec<BuildTarget> {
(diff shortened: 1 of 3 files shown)


boulder/src/env.rs
```diff
@@ -73,6 +80,27 @@ fn resolve_moss_root(is_root: bool, custom: Option<PathBuf>) -> Result<PathBuf,
...
+fn resolve_epoch() -> SystemTime {
+    // Propagate SOURCE_DATE_EPOCH if set
+    if let Ok(epoch_env) = env::var("SOURCE_DATE_EPOCH") {
+        if let Ok(time) = epoch_env.parse::<u64>() {
+            return SystemTime::UNIX_EPOCH.add(Duration::from_secs(time));
+        }
+    }
+
+    // If we are building from a git repo and have the git binary available to us then use the last commit timestamp
+    if let Ok(git_log) = Command::new("git").args(["log", "-1", "--format=\"%at\""]).output() {
```
ikeycode: I have a gut feeling this needs to explicitly set the working directory, does it work if you run boulder from a separate location to the stone.yaml ?

ReillyBrogan: You mean having the current directory be `/not/in/git/repo` and then invoking it as `boulder build /path/to/git/repo/stone.yaml`? No, in that case it would not correctly determine the git timestamp to use. It would be a bit fiddly to support that given that we're creating the `env` in a parent scope of `build` and `chroot` where we're guaranteed to have the path to the stone.yaml.

Is that something we want to support?

ReillyBrogan:
Alright I fucked around with this for a while and finally got it into a state where I am happy with it. SOURCE_DATE_EPOCH is now part of the recipe object which is logical and the directory containing the `stone.yaml` is used for the `git log` command.

ReillyBrogan:
After being introduced to our lord and savior the chrono crate I'm going to refactor this to use that instead.

boulder/src/recipe.rs
```diff
@@ -101,6 +110,35 @@ pub fn resolve_path(path: impl AsRef<Path>) -> Result<PathBuf, Error> {
...
+            if let Some(timestamp) = DateTime::from_timestamp(parsed, 0) {
+                return timestamp;
+            }
+        }
+    }
+
+    // If we are building from a git repo and have the git binary available to us then use the last commit timestamp
+    if let Ok(git_log) = Command::new("git")
+        .args(["log", "-1", "--format=\"%at\""])
+        .current_dir(path.parent().unwrap())
```
tarkah: Let's remove the unwrap. I know it's technically impossible since we've previously called `resolve_path` and it's a sanitized path, but this is still a potential panic if that implementation ever changes.

ReillyBrogan:
Changes:
- Removed unwrap
- Renamed `epoch` to `build_time`

tarkah (approved):

PR #239, opened 2024-05-09
boulder/moss: Support shell completions generation

ReillyBrogan:
This wires up clap_complete so that the `completions` subcommand of boulder and moss will now automatically generate shell completions when provided a shell as the single argument.

The second commit replaces the alias implementation in moss with one that works better with how Clap works with shell completions. Now commands like `moss ur` correctly show further completions as though the user had used `moss repo update`

Closes serpent-os/moss#220

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,6 +149,7 @@ version = "0.1.0"
 dependencies = [
  "chrono",
  "clap",
+ "clap_complete",
  "config",
  "container",
  "derive_more",
@@ -272,6 +273,15 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+]
+
+[[package]]
+name = "clap_complete"
+version = "4.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dd79504325bf38b10165b02e89b4347300f855f273c4cb30c4a3209e6583275e"
+dependencies = [
+ "clap",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


ikeycode:
OK I'm liking in principle but I don't actually know how I feel about matching on `None` as a magic value for completions?

ReillyBrogan:
> OK I'm liking in principle but I don't actually know how I feel about matching on `None` as a magic value for completions?

The reason I'm doing that is in order to support `moss li` as an alias for `moss list installed` we have to attach the command for `installed` to the root command object as `li` (but hidden so it doesn't appear in `--help`). Ordinarily when `moss list installed` is called we unpack the clap return as a subcommand `list` which has a further child `installed`. When we call it as `moss li` instead we just get the equivalent of the final `installed` object. So that's why the pattern matcher matches on `None` currently, if `moss li` is called then there's no further subcommand to extract out and route with so we need to use the additional context that's passed to the `install` handler.

An alternative would be to split up the "figure out what subcommand was called" logic from the "route to appropriate handler" logic which would probably make it look a bit cleaner.

ermo:
> An alternative would be to split up the "figure out what subcommand was called" logic from the "route to appropriate handler" logic which would probably make it look a bit cleaner.

Per Ikey's comments, this is probably a good idea, assuming you're up for it?

ikeycode:
unfortunately this died of death months ago so has been superseded in main

PR #191, opened 2024-03-26
Avoid regex in fnmatch crate

livingsilver94:
Rewrite the fnmatch crate to not use regex. It's fast (comparable to the wildmatch crate), tested and fuzzed. Should be usable already. Run `cargo bench` to compare it against wildmatch and regex.

Documentation is in code an in the crate-level README.md file. Launch `cargo doc --open --no-deps --document-private-items` for a decent layout.

PR #240, opened 2024-05-10
Make crates/triggers API suitable for usysconf

livingsilver94:
This is my attempt to decouple the current crates/triggers code, which was written specifically with moss in mind, to make it as generic as possible. Particularly, we want to re-write Solus' usysconf tool in Rust, leveraging the Serpent OS tech stack. usysconf works differently from "moss-triggers", yet we want to share as much code as possible. See this message for reference: https://github.com/serpent-os/usysconf-rs/issues/1#issuecomment-1714527980

NOTE: This PR sits on top of #191 and in fact pours its commit in the fnmatch-simplify branch. This is for simplicity, so that every PR only sees its relevant subset of commits. We'll eventually merge fn-simplify into main.


livingsilver94:
I think we should review (and hopefully accept) this PR before moving to usysconf. This PR in fact is the core API for both moss and usysconf. If there's something fishy here, there's something fishy everywhere.

livingsilver94:
I'm not sure this PR is still relevant today. I'm closing it; we can reopen if necessary.

PR #241, opened 2024-05-10
repo: Release improvements

ReillyBrogan:
The first commit adds an install script to the justfile that installs all required files for boulder and moss into a given prefix, suitable for use during building packages. Note that the release script assumes that serpent-os/moss#239 has already been merged as it attempts to generate completions. I felt that that particular commit made more sense combined with this one.

The second commit adds a script to generate a vendored tarball containing source code plus vendored dependencies, suitable for use building packages without networking. This tarball is completely reproducible. 

To prove reproducibility, you can check out this branch and run the script yourself. It should output the following checksum, assuming your zstd version is v1.5.6:
```
624463d1acf9eab911eba1d631ee3c6b33aaae300a98a359c2d8528753a7e220
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ target
 install
 
 **/db/**/test.db
+
+# Generated source archives
+*.tar.*
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 /target
 target
+/install
+install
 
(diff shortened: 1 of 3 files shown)


ermo (changes requested):
Needs rebase.

PR #244, opened 2024-05-14
boulder: feat/cli macro descriptions

tarkah:


ermo:
Did you test if this is [XDG](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) compliant already? 

This would enable us to install moss and boulder to the relevant XDG directories via the `just get-started` onboarding target.

FWIW, I've taken the liberty to force-push the main changes that include a README update and the justfile stuff. So you can pull/reclone and test `just get-started` with this branch.

ermo:
Force push was about ensuring that the justfile is the newest one from the main branch which now uses the default XDG_*_HOME dirs.

ermo:
Force pushed with newest build flags from main (faster build, smaller binaries).

ermo (approved):
I have now added a description in the moss README that shows how to get started with the current implementation.

I've also written up some design notes that we will want to use for proper XDG support, using a single code-path and some other tweaks, but this is really its own unit of work.

Hence I'm merging this as-is and leaving further improvements to future PRs.

PR #253, opened 2024-06-08
all: Use stdlib OnceLock for HTTP client

ReillyBrogan:
once_cell functionality was integrated into the stdlib in Rust 1.70.0. Switch to it so we can drop once_cell as a dependency.

While OnceLock is functionally similar to the thread-safe modules in once_cell the APIs are somewhat different. In this case the biggest difference is that you can't declare the function that creates the object when you declare the object itself, any caller that wants to use the object needs to provide the function to create it if it does not already exist. This is trivially worked around by wrapping the object in a function, which is what I did here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1263,7 +1263,6 @@ dependencies = [
  "libsqlite3-sys",
  "log",
  "nix",
- "once_cell",
  "rayon",
  "reqwest",
  "serde",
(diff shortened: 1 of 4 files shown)


ikeycode (approved):

ikeycode:
i luv less deps <3

PR #255, opened 2024-06-08
boulder: Add RUSTFLAGS, enable frame pointers by default

ReillyBrogan:
From the Rust docs:

> A space-separated list of custom flags to pass to all compiler invocations that Cargo performs. In contrast with `cargo rustc`, this is useful for passing a flag to _all_ compiler instances. (...) This string is split by whitespace

And define the default for that flag to be `-Cforce-frame-pointers`, which from the Rust docs force enables frame pointers. Note that setting this flag to `=false` or leaving it off does not necessarily mean that frame pointers will be removed as they can still be enabled by other means.

Though it is unlikely to be needed setting `frame-pointer: false` in the tuning section of a recipe will remove this flag.


ikeycode (approved):
beautifully simple.

PR #256, opened 2024-06-08
all: Cleanup dependencies

ReillyBrogan:
Changes:
- Pin all dependencies to the specific version for consistency (exact versions will be managed by dependabot/renovate)
- Move all specific version pinning to the top-level cargo.toml
- Cleanup unused dependencies throughout the project


ikeycode:
Looks like this broke unixepoch? o_O

ReillyBrogan:
> Looks like this broke unixepoch? o_O

Fixed

ikeycode (approved):

PR #258, opened 2024-06-11
boulder: Implement recipe bump support

### LINKED ISSUE #208: boulder: Add `bump` subcommand that parses the current manifest buildrelease and bumps it by one
Motivation:

Sometimes, it's nice to be able to bump a recipe for a rebuild. However, if that recipe is unchanged, it makes more sense for this behaviour to be done by the (already existing) `boulder build -b<existing buildrelease>+1` functionality.

Benefit:

This saves us a heap of YAML-related trouble in figuring out a way to keep "nice" formatting, because this is a purely read-only operation that just reads the existing manifest for the existing buildrelease and then adds one to that.

This in turns makes it simple to automate rebuilds and commit them via a convenient and repeatable `autobuild`-driven script solution.

joebonrichie:
Increments the release of a recipe by one or optionally set release to any value.

### DIFF
--- a/boulder/src/cli/recipe.rs
+++ b/boulder/src/cli/recipe.rs
@@ -34,6 +34,23 @@ pub struct Command {
 
 #[derive(Debug, clap::Subcommand)]
 pub enum Subcommand {
+    #[command(about = "Bump a recipe's release")]
+    Bump {
+        #[arg(
+            short,
+            long,
+            default_value = "./stone.yaml",
+            help = "Location of the recipe file to update"
+        )]
+        recipe: PathBuf,
+        #[arg(
+            short = 'n',
+            long,
+            required = false,
+            help = "Set release to a specific number instead of incrementing by one"
+        )]
+        release: Option<u64>,
+    },
     #[command(about = "Create skeletal stone.yaml recipe from source archive URIs")]
     New {
         #[arg(
@@ -91,6 +108,7 @@ fn parse_upstream(s: &str) -> Result<Upstream, String> {
 
 pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     match command.subcommand {
+        Subcommand::Bump { recipe, release } => bump(recipe, release),
         Subcommand::New { output, upstreams } => new(output, upstreams),
         Subcommand::Update {
             recipe,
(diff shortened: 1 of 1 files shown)


boulder/src/cli/recipe.rs
```diff
@@ -102,6 +120,38 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
...
+    let prev = parsed.source.release;
+    let next = release.unwrap_or(parsed.source.release + 1);
+    let mut updater = yaml::Updater::new();
+    updater.update_value(next, |root| root / "release");
+
+    // Apply updates
+    let updated = updater.apply(input);
+
+    let recipe = recipe.expect("checked above");
+    fs::write(&recipe, updated.as_bytes()).map_err(Error::Write)?;
```
tarkah: recipe_path is already available

boulder/src/cli/recipe.rs
```diff
@@ -102,6 +120,37 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     }
 }
 
+fn bump(recipe: Option<PathBuf>, release: Option<u64>) -> Result<(), Error> {
+    let Some(recipe_path) = &recipe else {
+        panic!("Can't resolve recipe path: {}", recipe.unwrap().display())
```
tarkah: Why is `recipe` optionally defined in the clap args? Make it non-optional and then we won't have to deal w/ this impossible state. `clap` provides a much nicer UX for missing values on CLI than we can print ourselves.

boulder/src/cli/recipe.rs
```diff
@@ -34,6 +34,23 @@ pub struct Command {
 
 #[derive(Debug, clap::Subcommand)]
 pub enum Subcommand {
+    #[command(about = "Bump a recipe's release")]
+    Bump {
+        #[arg(
+            short,
+            long,
+            default_value = "./stone.yaml",
+            help = "Location to output generated build recipe"
```
tarkah: Nit: Something like this? 

```suggestion
            help = "Location of the recipe file to update"
```

boulder/src/cli/recipe.rs
tarkah: Also, we provide a `default_value` so if not provided by the user, it'll default to `./stone.yml`. So the optional is very much not needed.

tarkah (approved):

Issue #168, opened 2024-03-07
-h and --help often show different information


joebonrichie:
e.g.

```
$ ~/sos/moss/target/release/boulder profile add -h
Add a new profile

Usage: boulder profile add [OPTIONS] --repo <REPOS> <NAME>

Arguments:
  <NAME>  profile name

Options:
      --cache-dir <CACHE_DIR>    
  -r, --repo <REPOS>             repository to add to profile, can be passed multiple times
      --config-dir <CONFIG_DIR>  
      --data-dir <DATA_DIR>      
      --moss-root <MOSS_ROOT>    
  -h, --help                     Print help (see more with '--help')
```
vs
```
$ ~/sos/moss/target/release/boulder profile add --help
Add a new profile

Usage: boulder profile add [OPTIONS] --repo <REPOS> <NAME>

Arguments:
  <NAME>
          profile name

Options:
      --cache-dir <CACHE_DIR>
          

  -r, --repo <REPOS>
          repository to add to profile
          
          Example: --repo name=volatile,uri=https://dev.serpentos.com/volatile/x86_64/stone.index,priority=100

      --config-dir <CONFIG_DIR>
          

      --data-dir <DATA_DIR>
          

      --moss-root <MOSS_ROOT>
          

  -h, --help
          Print help (see a summary with '-h')
```

Can see that `--help` provides an example for options but `-h` does not, is this expected behaviour?

livingsilver94:
I agree with you that `-h` and `--help` should just be the short and the long forms of the same flag, but it looks like it's wanted: https://github.com/serpent-os/moss/blob/main/boulder/src/cli/profile.rs#L36

joebonrichie:
Alright i'll close for now then.

tarkah:
> I agree with you that `-h` and `--help` should just be the short and the long forms of the same flag, but it looks like it's wanted: https://github.com/serpent-os/moss/blob/main/boulder/src/cli/profile.rs#L36

I don't think there's been a discussion yet around this, so it's a very valid point! This is the behavior provided by clap that I opted into, but we can definitely omit long_help and just use help if we want the same messages everywhere.

I agree it's a little confusing since users won't know they're getting a condensed vs full help. I think we just show them the whole thing, that's more sane I think.

@joebonrichie feel free to reopen so we can track. Would be good to get the opinion of our resident UX expert @ermo

ermo:
Show the whole thing on `-h/--help` IMO.

[...]

PR #266, opened 2024-07-07
Print help when argument required but not provided

dylanchang44:
PR related to #168  discussion.
When additional argument needed, print short help instead of default clap error.
Align for both moss and boulder.

#Before
$boulder profile add
```
error: the following required arguments were not provided:
  --repo <REPOS>
  <NAME>

Usage: boulder profile add --repo <REPOS> <NAME>

For more information, try '--help'
```

#After
$boulder profile add
```
Add a new profile

Usage: boulder profile add [OPTIONS] --repo <REPOS> <NAME>

Arguments:
  <NAME>  profile name

Options:
      --cache-dir <CACHE_DIR>    
  -r, --repo <REPOS>             repository to add to profile, can be passed multiple times
      --config-dir <CONFIG_DIR>  
      --data-dir <DATA_DIR>      
      --moss-root <MOSS_ROOT>    
  -h, --help                     Print help (see more with '--help')
```

### DIFF
--- a/boulder/src/cli/profile.rs
+++ b/boulder/src/cli/profile.rs
@@ -23,7 +23,7 @@ pub struct Command {
 pub enum Subcommand {
     #[command(about = "List profiles")]
     List,
-    #[command(about = "Add a new profile")]
+    #[command(about = "Add a new profile", arg_required_else_help(true))]
     Add {
         #[arg(help = "profile name")]
         name: String,
--- a/boulder/src/cli/recipe.rs
+++ b/boulder/src/cli/recipe.rs
@@ -51,7 +51,10 @@ pub enum Subcommand {
         )]
         release: Option<u64>,
     },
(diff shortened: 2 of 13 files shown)


ReillyBrogan:
Please take a look at the commit messages we have on [main](https://github.com/serpent-os/moss/commits/main/) and edit your commits so that they match that format. You also need to squash all of your commits into one here, the only logical separation I can see would be if you split the boulder and the moss changes to be separate commits but it's also fine if they're just one commit. We don't need additional commits that are just typo/lint fixes though.

tarkah:
Thanks for the PR! 

Hmm, I find the first error output more helpful since it explicitly tells the user what is missing. It's not clear by just printing out help what went wrong.

dylanchang44:
Apologies for not following the format, and will squash the commits next time.
If the first error output is preferred, then I think this PR can be withdrawn.

PR #268, opened 2024-07-09
all: Version improvements

ReillyBrogan:
The first and second commits centralize version handling to the top-level Cargo.toml file and add a MSRV respectively (one of the build deps has a MSRV on 1.78 which is why it's so high comparatively).

The third commit adds a new serpent_buildinfo crate which processes git info, version, and build timestamps during the build and exposes a simple API for boulder and moss to use. This uses the git CLI directly and should work with git workspaces. It should also work if a .git directory is not present (intended to work with serpent-os/moss#241) at which point it will only capture the version and build timestamp.

Since the serpent_buildinfo crate will be rebuilt on virtually all code changes I tried to ensure that it builds quickly. `cargo::rerun-if-changed` instructions are setup to rebuild it on any changes to source code directories (non-source code directories and files should be skipped).

It also respects SOURCE_DATE_EPOCH and will use that for the build timestamp if present.


ikeycode (approved):
ta :)

PR #275, opened 2024-07-15
Allow multiple arguments to %cargo_install

Staudey:
This allows multiple arguments to the `%cargo_install` macro, in case there are several binaries to install

See also https://github.com/getsolus/ypkg/pull/90


ermo (approved):
LGTM, thanks!

PR #276, opened 2024-07-15
boulder: Initial support for rust pgo

joebonrichie:
https://doc.rust-lang.org/rustc/profile-guided-optimization.html

* caveats: doesn't work with gcc toolchain due to it relying on `llvm_profile_merge`


ermo:
Might as well get this in for more general testing.

ermo (approved):

PR #278, opened 2024-07-16
boulder: Use shared moss root in boulder cache

### LINKED ISSUE #271: boulder: Do not use `/.moss` store
When invoking `boulder build` as root it uses `/.moss` for the store, rather than a *private* moss store.
This seems bad.

tarkah:
We don't want to default to using the system moss root for boulder, as we want paranoid levels of protection of the system root (boulder shouldn't be able to
accidentally clobber it due to a bug).

Use a shared moss root in the boulder cache directory instead.

Resolves #271


ikeycode (approved):
Took me a while to grok but I think it works >_>

PR #280, opened 2024-07-17
Rejigger cargo manifest and config

ermo:
With these changes, `just get-started` works for me again on Fedora.

Read the comments in `./.cargo/config.toml` file for my suggestions on how we might be able to handle specifying extra flags when cargo detects that it is running on a Serpent-native rust toolchain.


Cargo.toml
```diff
@@ -54,25 +55,4 @@ url = { version = "2.5.2", features = ["serde"] }
...
-# We want people who use the onboarding steps to get a nice compromise
-# between fast compilation and fast runtime, but with checks in place
-# and full backtraces. Hyperfine tests shows opt-level = 1 to be a good
-# compromise between compile speed and runtime speed.
-[profile.onboarding]
-inherits = "dev"
-opt-level = 1
-lto = "thin"
-debug = true
-strip = "none"
```
ReillyBrogan: I don't think this is a good idea. Cargo.toml is for project or workspace configuration, while config.toml is for cargo configuration. The intent is that someone can override the config.toml at build time and get a different set of behavior.

boulder/build.rs
```diff
@@ -0,0 +1,15 @@
+use etc_os_release::OsRelease;
+use std::error::Error;
+use std::path::PathBuf;
+
+/// Set cargo::rustc-cfg=os_release_id="whatever" when /etc/os-release warrants it
+fn main() -> Result<(), Box<dyn Error>> {
+    // only recompile when necessary
+    let top_level = PathBuf::from("..").canonicalize()?;
+    println!("cargo::rerun-if-changed={}/build.rs", top_level.display());
```
ReillyBrogan: These lines were only necessary since you were using a symlink for build.rs previously. Cargo already by default considers changes within the directory being built to invalidate cache results, so this is unnecessary.

.cargo/config.toml
```diff
@@ -1,22 +1,71 @@
...
+#   we have ABI reports. Upstream hasn't switched to it yet by default due to
+#   stable distros not having new enough tools, but that doesn't matter for us
+#   [1]:  https://doc.rust-lang.org/rustc/symbol-mangling/v0.html
+#
+
+# NB: os_release patterns need to be added to both target configs for this to
+#     work...
+#
+# The Solus toolchain supports zstd debug sections currently (Serpent doesn't)
+[target.'cfg(any(os_release_id = "solus"))']
```
ReillyBrogan: I think it would be better to remove this and have build.rs emit rustflag configurations instead. I think your goal here is that these configuration options would take effect for compilation of all crates, but that's not actually going to work since project dependencies are built BEFORE execution of the build.rs of a given project/crate. Since `link-arg`s are only used at link-time putting them in build.rs is perfectly fine since `boulder` and `moss` are where the linking happens anyway.

ermo:
Converting to draft as this clearly needs more work.

PR #281, opened 2024-07-17
boulder: Add support for compressman option

joebonrichie:
Compresses any man/info pages with zstd. A tradeoff between slightly bigger .stone files vs smaller installed size.

Tested 
- `nano` (contains man and info files)
- `xz` (contains lots of man symlinks)


boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
...
+    }
+
+    pub fn compress_file_zstd(path: PathBuf) -> Result<PathBuf, BoxError> {
+        extern crate zstd;
+        use std::fs::File;
+        use std::io::{BufReader, BufWriter, Write};
+        use zstd::stream::copy_encode;
+
+        let output_path = PathBuf::from(format!("{}.zst", path.display()));
+        let input = File::create(&output_path).unwrap();
```
ReillyBrogan: `.unwrap()` will panic if the `Result` has an object, which is not a good idea and you should avoid using it.

Instead, you can use `?` (see [here](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html)) to automatically handle unwrap and handle errors. If a function returns an error it will automatically unwrap it if it's not an error and continue, and if it is an error it will propagate it upwards. This requires that the function with `?` returns a `Result`, but you're already doing that.

So
```
        let input = File::create(&output_path).unwrap();
```
would become
```
        let input = File::create(&output_path)?;
```

Use this pattern every where you are currently doing an unwrap and profit.

boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
...
+
+        let mut reader = BufReader::new(File::open(path).unwrap());
```
ReillyBrogan:
```
        let mut reader = BufReader::new(File::open(path)?);
```

boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
...
+        }
+    } else {
+        return Ok(Decision::NextHandler.into());
+    }
+
+    pub fn compress_file_zstd(path: PathBuf) -> Result<PathBuf, BoxError> {
+        extern crate zstd;
+        use std::fs::File;
+        use std::io::{BufReader, BufWriter, Write};
+        use zstd::stream::copy_encode;
```
ReillyBrogan: I assume you copied this section from somewhere:
- Don't use `extern crate zstd;`. All it's doing is allowing you to use `copy_encode()` without the zstd namespace, but that's not actually all that useful. You should just be able to use `zstd::copy_encode()` after removing it (and having the correct imports)
- Move the `use` statements above and merge with the existing imports

boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
...
+                let compressed_file = compress_file_zstd(PathBuf::from(format!(
+                    "{}/{}",
+                    &parent_path,
+                    fs::read_link(&file_path)?.display()
+                )))
+                .unwrap_or_default();
+                let _ = bucket.paths.install().guest.join(compressed_file);
+            }
+
+            symlink(format!("{}.zst", fs::read_link(&file_path)?.display()), &new_link).unwrap_or_else(|why| {
```
livingsilver94: This is pretty convoluted. Since `fs::read_link` returns a `Result` contaning a `PathBuf`, can you just do `fs::read_link(&file_path)?.set_extension("zst")` ?
 
[Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=018c6a785cdeaac4924d2a869b28315d) here.

boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
 
     Ok(Decision::NextHandler.into())
 }
+
+pub fn compressman(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, BoxError> {
+    let file_path = info.path.clone().into_os_string().into_string().unwrap_or_default();
```
livingsilver94: Do you need this to be a String, instead of a Path or PathBuf? I see that below you're calling `contains`, `ends_with` and other string operations like these.

As an example, the `contains` function can be replicated by calling `components`, which is an iterator, and then by checking if the iterator contains `man`. `ends_with` could be checking whether the extension of `file_path.file_name()` (which returns the last component of the path) or the full name is what you expect.

boulder/src/package/analysis/handler.rs
```diff
@@ -134,3 +134,82 @@ pub fn cmake(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, Bo
...
+    if (file_path.contains("man") && file_path.ends_with(|c| ('1'..'9').contains(&c)))
+        || (file_path.contains("info") && file_path.ends_with(".info"))
+    {
+        /* If we have a man/info symlink update the link to the compressed file */
+        if info.path.is_symlink() {
+            let parent_path = info.path.parent().unwrap().display();
+            let new_original = format!("{}/{}.zst", &parent_path, fs::read_link(&file_path)?.display());
+            let new_link = format!("{}.zst", &file_path);
+
+            /*
```
livingsilver94: C-like comments are meh in my opinion. Double slashes ftw.

livingsilver94 (changes requested):
Although I'm not a Rust expert, let's try to make this code more idiomatic by using the Path/PathBuf structs.

boulder/src/package/analysis/handler.rs
joebonrichie: set_extension returns a bool so no longer a one liner

joebonrichie: `with_extension()` is a no-go as it replaces 1..9 with the extension.

joebonrichie:
Should be good now

joebonrichie:
:ping_pong:

ermo (changes requested):
Needs a rebase. I am aware that this is inconvenient and that it's been sitting for _way_ too long in the queue.

ermo (changes requested):
Needs rebase sadly.

joebonrichie:
replaced by #475

PR #282, opened 2024-07-18
moss: Fix duplicate inserts

ReillyBrogan:
This fixes the following error:

```
Error: build: build recipe: root: moss client: repository manager: meta db: diesel: UNIQUE constraint failed: meta_licenses.package, meta_licenses.license
```

It appears that the section that generates metadata for inserting into the database is generating multiple entries with the same id and value (these tables all have primary keys on the id and values). Perhaps the better solution is to de-duplicate the array before inserting it or figure out why it's creating duplicate items in the first place, but this gets things building on my machine at least.


tarkah (approved):

tarkah:
Perfect thanks!

PR #283, opened 2024-07-18
boulder/moss: Reduce binary size by reducing dependency features

ReillyBrogan:
I used cargo-bloat to find out what some of the biggest offenders with regards to binary size were and found that reqwest and regex could both be trimmed a bit by disabling some features that we don't need.

For regex I switched it to only have the `std` feature enabled. This disables some of the performance features and unicode support but as far as I can tell we don't use the unicode support and the existing regexes are in non-performance-critical sections. This reduces boulder binary size with the `release` profile from `16.9MiB` -> `15.2MiB`.

For reqwest I switched to using `native-tls` over `rustls`. Since this links against the system openssl the binary size is further reduced from `15.2MiB` -> `13.1MiB`.

Final binary sizes when this patch is applied to the experimental Solus package (which already links against system libzstd and sqlite3) are:
- Boulder: 11.8MiB -> 9.2MiB (-22%)
- Moss: 10.7MiB -> 8.1MiB (-24%)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -374,6 +374,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+ "terminal_size",
 ]
 
 [[package]]
@@ -443,6 +444,16 @@ name = "convert_case"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
+
+[[package]]
+name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
 
 [[package]]
 name = "core-foundation-sys"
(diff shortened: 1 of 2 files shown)


ikeycode:
I'd whole heartedly understand why this is desirable. However...

I'd sooner not rely on openssl at all, and even our curl package in serpent uses it. The other dependencies would introduce issues in volatile whereby we could dep-break the repo or images, rather than controlled isolated binaries.

Yes, it's an annoying approach but having an almost entirely static build eliminates a plethora of LD issues and makes iteration less problematic (consider the solbuild + libgit2 issues we've known to love and cherish)

Ideally we should trim fat where feasible and nuke lardy deps like petgraph and Co, and use the "packaging" moss target which already yields a similar binary size to the one you've provided.

tarkah:
@ReillyBrogan can we remove the `rustls` change so we can get these other fixes in? They're still very material.

ermo:
> @ReillyBrogan can we remove the `rustls` change so we can get the `regex` change in? It's still very material on it's own.

Is this still an issue @tarkah ?

As soon as we're happy with infra alpha2 (and you are under less pressure re. your move and such), I'd love to have you and Jonas look over the remaining alpha1 and milestone-less PRs with me so we can plan os-tools alpha2 and what we want in for that larger rework.

livingsilver94:
Can we at least merge the `regex`-stripping part? I absolutely agree with Reilly: we're not using regexes in any critical path and a little reduction in final size is welcome. At the moment `moss` goes from ~20 MiB to ~19 MiB.

PR #293, opened 2024-07-25
Boulder: hook up rust optimization flags & adjust cargo_build macro

joebonrichie:
Let us control rust optimization options via our flags.

Also fixes the buggy cargo_build macro not passing intended --release optimization options.


ikeycode (approved):
very nicely done

PR #302, opened 2024-08-04
Split ContainerError::Nix into more specific variants

jplatte:
I'm getting this weird error when running moss from a chroot:

```
Error: sync: client: postblit: container: exited with failure: EINVAL: Invalid argument
```

This change helped me narrow down what's going wrong (it's the call to `pivotroot`).


ermo:
Nice!

ikeycode (approved):
definitely needed!

PR #306, opened 2024-08-05
crates/container: Ensure localhost networking is setup

joebonrichie:
This fixes an issue where the sccache daemon could not start with networking disabled. Issue was not noticed as most rust projects do not vendor their dependencies.


ermo:
If we do this as-is using `ip`, we'll need to rundep on it in the recipe for it?

joebonrichie:
> If we do this as-is using `ip`, we'll need to rundep on it in the recipe for it?

it's done before pivot_root to avoid forcing iproute2 into the container

ikeycode (approved):
i hate that we have to do this, but in the absence of a suitable alternative whats the point in it hanging around?

PR #308, opened 2024-08-07
ci: Use more sophisticated cache action

jplatte:
See https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details for details.


jplatte:
If you agree with this, I'm happy to make PRs for the other Rust repos in serpent-os.

ikeycode (approved):

jplatte:
Wow, this had a much more positive effect than I expected!

PR #314, opened 2024-08-10
Moar CI updates

jplatte:
I'm not completely certain w.r.t. caches of the nightly thing, but I guess it will be pretty obvious with the next nightly (currently the times are pretty stable at slightly above 5min).


ikeycode (approved):
Only one way to find out!

jplatte:
Didn't slow down the build! :)

PR #316, opened 2024-08-13
moss/boulder: Add support for font providers

joebonrichie:
Reads FAMILY_NAME from OpenType fonts in order to provide a font provider.

e.g. `moss it 'font(Noto Sans Tamil Condensed Thin)'`


ikeycode:
This will temporarily require support in `libmoss` for the infra (out as far as MetaDB) but we need to verify if we can even handle this without a format bump. Perhaps an overarching goal for our first format bump is needed, including restoration of timestamps.

PR #317, opened 2024-08-13
boulder: Hook up python providers

joebonrichie:
Uses mailparse to parse the RFC style package information from .dist-info or .egg-info directories and provides a python provider from the name.

Based off `pip3 show`


ikeycode (approved):
🫨

PR #325, opened 2024-08-16
moss: Fix sync transaction & use pinned providers

tarkah:
### Issue

Transaction resolution was borked in `sync` due to the fact we were explicitly adding transitive packages to the transaction.

This is incorrect since transitive packages should only get added to a transaction via provider resolution from a dependent explicit package. There are two issues from doing this:

- The explicitly added transitive might be a different version than what's resolved from the dependent explicit during provider resolution. Both versions would end up in the final TX.
- Transitives of the explicitly added transitive might be orphaned if the newer explicitly added transitive version no longer has that dependency. Then that orphaned transitive got added because we added all original transitives from the old state.

Even if we solve the above by removing orphaned packages from the TX, the _wrong_ package would get removed because the transitive with an edge to the explicit was the _resolved_ transitive, not the explicitly added one.

The reason we explicitly added transitives was to ensure sync updates to transitive packages were picked up in the final state. This was needed because TX provider resolution will first resolve against selections and installed packages before available packages.

### Solution

Allow a TX to resolve providers first based on a list of "pinned" packages. During sync, we can define this pinned list of providers as "packages that have a sync update". Then we _only_ add explicit packages to the TX. When provider resolution is done on those explicits dependencies, if it matches one of the pinned packages, it will use that first before then falling back to selections -> installed -> available.

This results in a transaction that cannot have orphans because it only resolved from explicit packages while also resulting in a TX that includes updated transitives if they were part of the pinned provider list.

### DIFF
--- a/moss/src/registry/transaction.rs
+++ b/moss/src/registry/transaction.rs
@@ -14,6 +14,10 @@ enum ProviderFilter {
     /// Filter the lookup to current selection scope
     Selections(Provider),
 
+    /// Use a pinned package if one exists
+    /// for this provider
+    Pinned(Provider),
+
     // Look beyond installed/selections
     All(Provider),
 }
@@ -31,6 +35,11 @@ pub struct Transaction<'a> {
 
     // unique set of package ids
     packages: Dag<package::Id>,
+
+    /// packages which are always considered first
+    /// during [`ProviderFilter::Pinned`] but
+    /// aren't part of `packages` DAG
+    pinned_providers: Vec<package::Id>,
 }
 
 /// Construct a new Transaction wrapped around the underlying Registry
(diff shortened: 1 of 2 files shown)


moss/src/registry/transaction.rs
```diff
@@ -40,6 +49,7 @@ pub(super) fn new(registry: &Registry) -> Result<Transaction<'_>, Error> {
     Ok(Transaction {
         registry,
         packages: Dag::default(),
+        pinned_providers: vec![],
```
ikeycode, line 52: Smart! Total inverse of my local works which used an occlusion mechanism, same result ofc!

ikeycode (approved):

PR #326, opened 2024-08-16
moss: Allow sync on installed pkg missing from repos

### LINKED ISSUE #307: moss: locally installed pkgs which no longer have a repo origin block sync
Screenshot from 2024-08-06 00-05-48

`key-rack` was originally installed via a user configured local repo. If `key-rack` no longer exists in the local repo index or the local remove is removed it blocks `moss sync`

tarkah:
We shouldn't error out if an installed package isn't available in a remote repo. Instead, we just keep
the package as is.

Resolves #307


ikeycode (approved):
muchos gracias, i know this was a pain point.

PR #329, opened 2024-08-16
moss: Ignore SIGINT during blit

### LINKED ISSUE #310: moss: interrupting a blitting operation is possible
Currently you can SIGINT during a blitting operation leaving to a completely borked moss state.

This should be generally be recoverable with `moss state` given the system is not too badly mangled.

- moss should prevent SIGINT/SIGKILL/SIGSTOP etc during a blitting operation
- moss should integrate with [systemd's inhibitor ](https://www.freedesktop.org/wiki/Software/systemd/inhibit/) to prevent the  system from sleeping/shutting down etc during a moss operation.

tarkah:
Partially handles #310

### DIFF
--- a/moss/src/client/mod.rs
+++ b/moss/src/client/mod.rs
@@ -38,9 +38,9 @@ use self::verify::verify;
 use crate::{
     db, environment, installation, package,
     registry::plugin::{self, Plugin},
-    repository, runtime,
+    repository, runtime, signal,
     state::{self, Selection},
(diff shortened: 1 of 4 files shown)


ikeycode (approved):
merging for expediency sake but ofc will test locally now =)

tarkah:
I added a `sleep(10s)` after the caller of `new_state` to test that Ctrl-c doesn't kill during blit, but it does after the blit

PR #336, opened 2024-08-19
moss: vfs: Cache expensive path calc / allocations

tarkah:
Basically fix `path() path() path() path() path()` being called 100x times. Cache it on first invocation by wrapping `BlitFitle` and calling these expensive methods only once.

Wall time went from 14.4s -> 600ms xD


ikeycode (approved):
brave dude strikes again

PR #337, opened 2024-08-19
boulder: Add support for automatic python dependencies

joebonrichie:
Compiles get-py-deps.py into the binary then executes at runtime using `python3 -c`

get-py-deps.py uses `importlib.metadata` from the python std to get a list of dependencies from either .egg-info or .dist-info paths. Then uses APIs from packaging to sanitize those dependencies.

By default "extra" dependencies are ignored and it is up to the packager's discrection to optionally include them.
Dependencies that depend on environmental markers e.g. Windows, python3 > $version will be correctly evaluated and included if necessary.

Depends on https://github.com/serpent-os/recipes/pull/359

### DIFF
--- a/boulder/src/package/analysis/handler.rs
+++ b/boulder/src/package/analysis/handler.rs
@@ -129,6 +129,7 @@ pub fn python(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, B
         return Ok(Decision::NextHandler.into());
     }
 
+    /* Add python providers name first */
     let data = std::fs::read(&info.path)?;
     let mail = parse_mail(&data)?;
     let python_name = mail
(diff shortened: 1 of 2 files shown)


boulder/src/package/analysis/handler.rs
```diff
@@ -138,6 +142,24 @@ pub fn python(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, B
...
 
+    /* Now parse dependencies */
+    let dist_path = info.path.parent().unwrap();
+    let find_deps_script = &bucket.paths.scripts().guest.join("get-py-deps.py");
+
+    let output = Command::new("/usr/bin/python3")
+        .arg(find_deps_script)
+        .arg(dist_path)
+        .stdout(Stdio::piped())
+        .output()?;
```
tarkah: I think we should compile this script into boulder so there's a guaranteed contract of what behavior we should expect here. Otherwise what if the file is missing or malformed.

We can place the .py alongside this module and `include_string!` it and pass that as an arg to be executed. You can template the `dist_path` arg and replace it before passing it to python.

We can then entirely remove scripts dir and having to ship it to the end user / expose it to the container.

ikeycode: That's a super nifty idea and creates less potential for hostside borkage ("drop in binary")

joebonrichie:
I guess we should try this then see where the shortcomings are, then we'll find out if we need a way to break autodeps or not.

ikeycode:
waiting for 359 to complete then will deploy

PR #340, opened 2024-08-20
moss/boulder: Add perl providers

joebonrichie:
Reads the "package" from .pm files to grok the provider.

### DIFF
--- a/boulder/src/package/analysis/handler.rs
+++ b/boulder/src/package/analysis/handler.rs
@@ -1,3 +1,5 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader};
 use std::{path::PathBuf, process::Command};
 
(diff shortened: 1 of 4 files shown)


boulder/src/package/analysis/handler.rs
```diff
@@ -64,6 +64,40 @@ pub fn binary(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Respons
     Ok(Decision::NextHandler.into())
 }
 
+pub fn perl(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response, BoxError> {
```
ermo, line 68: This would probably benefit from documentation showing what a typical perl module line that can be parsed actually looks like.

PR #341, opened 2024-08-20
boulder: Update cargo macros

ReillyBrogan:
Instead of building packages in debug profile (which requires workarounds for at least every package that uses rust-embed) build in release profile but with environmental variable overrides for the settings we need in order for RUSTFLAGS to take precedence.

This was tested by building several packages with different LTO profiles and verifying that they built as expected.

Also, add a few helper definitions for the artifact directory and the cargo options so that they can be used in other places.

### DIFF
--- a/boulder/data/macros/actions/cargo.yaml
+++ b/boulder/data/macros/actions/cargo.yaml
@@ -1,4 +1,14 @@
 actions:
+    - cargo_set_environment:
+        description: Set environmental variables for Cargo build
+        command: |
+            CARGO_BUILD_DEP_INFO_BASEDIR="%(workdir)"; export CARGO_BUILD_DEP_INFO_BASEDIR;
+            CARGO_NET_RETRY=5; export CARGO_NET_RETRY;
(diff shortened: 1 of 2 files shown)


tarkah (approved):
lol how did we miss this? I guess we always passed `--profile` for moss / boulder, but yeah `release` _needs_ to be the default :man_facepalming: 

Extra changes look good :+1:

ReillyBrogan:
It was present earlier, but was removed by @joebonrichie since he wanted the LTO configuration in the Boulder recipe to overwrite whatever was present in the Cargo.toml, but if LTO was enabled by Cargo.toml it would add flags that couldn't be disabled by RUSTFLAGS. The solution at the time was to build everything in debug profile, but this is a better approach. This overwrites any LTO configuration in the Cargo.toml so that it can be overrode by LTO flags added to RUSTFLAGS.

ikeycode (approved):
deploying new tooling this morning so lets give it a trial run

PR #343, opened 2024-08-22
moss: Add ability to disable repositories

tarkah:
image

### DIFF
--- a/boulder/src/cli/profile.rs
+++ b/boulder/src/cli/profile.rs
@@ -71,6 +71,7 @@ fn parse_repository(s: &str) -> Result<(repository::Id, Repository), String> {
             description: String::default(),
             uri,
             priority: repository::Priority::new(priority),
+            active: true,
         },
     ))
 }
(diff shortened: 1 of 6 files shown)


ikeycode (commented):
hold on, won't this require us overwriting a vendor shipped repo configuration in `/usr` ... ? which would be overwritten by updates..?

tarkah:
> hold on, won't this require us overwriting a vendor shipped repo configuration in `/usr` ... ? which would be overwritten by updates..?

We only write updates to admin /etc or user config dir

tarkah:
@ikeycode I've validated this works as expected:

- If we ship default repo / `volatile` to `/usr/share/moss/repo.d`, it shows up in `moss lr`
- User wants to "override" this / disable the repo: `moss dr volatile`
- Moss writes the overwrite to _admin_ at `/etc/moss/repo.d/volatile.yaml` w/ `active: false`
- `moss lr` shows it disabled, since it always merges admin configs _over_ vendor configs

ikeycode (approved):
<3

PR #345, opened 2024-08-22
boulder: Add versioned python providers

joebonrichie:
Add a versioned in addition to the generic provider. This will help with autodep breaking in the case of automatic dependencies.

In the future if we provide more than one python interpreter we will need to drop the generic provider for the "non-main" python candidate.

### DIFF
--- a/boulder/src/package/analysis/handler.rs
+++ b/boulder/src/package/analysis/handler.rs
@@ -1,4 +1,7 @@
-use std::{path::PathBuf, process::Command};
+use std::{
+    path::PathBuf,
+    process::{Command, Stdio},
+};
 
 use moss::{dependency, Dependency, Provider};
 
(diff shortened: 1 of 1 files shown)


boulder/src/package/analysis/handler.rs
```diff
@@ -133,11 +136,25 @@ pub fn python(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, B
         .get_first_value("Name")
         .unwrap_or_else(|| panic!("Failed to parse {}", info.file_name()));
 
+    /* Insert generic provider */
     bucket.providers.insert(Provider {
         kind: dependency::Kind::Python,
         name: python_name.to_string(),
     });
 
+    let output = Command::new("/usr/bin/python3")
```
ikeycode, line 145: In future it could be worth caching this for the run?

ikeycode (approved):

PR #347, opened 2024-08-23
boulder: Inherit root build if profile is missing phase

### LINKED ISSUE #338: boulder: profiles does not inherit from root key where profile is empty
See https://github.com/serpent-os/recipes/pull/244 as an example

e.g.
```yaml
emul32      : yes
profiles    :
    - emul32:
        setup: |
            do emul32 specific thing
setup       : |
    do thing
build       : |
    build
install     : |
    install
```

the emul32 phases for build and install do not get executed here, only profile; due to the addition of the emul32 specific setup phase.

tarkah:
Resolves #338

### DIFF
--- a/boulder/src/build/root.rs
+++ b/boulder/src/build/root.rs
@@ -107,8 +107,26 @@ fn packages(builder: &Builder) -> Vec<&str> {
         packages.extend(CCACHE_PACKAGES);
     }
 
-    packages.extend(builder.recipe.parsed.build.build_deps.iter().map(String::as_str));
-    packages.extend(builder.recipe.parsed.build.check_deps.iter().map(String::as_str));
+    packages.extend(
+        builder.recipe.parsed.build.build_deps.iter().map(String::as_str).chain(
+            builder
+                .recipe
+                .parsed
+                .profiles
+                .iter()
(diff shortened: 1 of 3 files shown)


boulder/src/build/root.rs
```diff
@@ -107,8 +107,26 @@ fn packages(builder: &Builder) -> Vec<&str> {
...
+    packages.extend(
+        builder.recipe.parsed.build.check_deps.iter().map(String::as_str).chain(
+            builder
+                .recipe
+                .parsed
+                .profiles
+                .iter()
+                .flat_map(|kv| kv.value.check_deps.iter().map(String::as_str)),
+        ),
+    );
```
tarkah, line 129: @ikeycode This is a potential recipe smell. Should profile overrides be allowed to define their own build / check deps? They're installed builder-wide anyways so it's no different than just defining them on the root.

Should we disable defining these two items for profile overrides?

ikeycode: I think for now we'll keep as is, because it might be "nice" to whack builddeps under a local. I'll see how I feel over time but yes there's a chance of reversion ^^

ikeycode (approved):

PR #349, opened 2024-08-25
boulder: cfg! compat_dlang_emul_both -> compat_dlang_emul_flush

joebonrichie:
emul32 libs will now depend on 386 variants of the provided emul32 libs instead of x86.

Part of #223.


ikeycode (approved):
🎊

PR #350, opened 2024-08-25
moss+boulder: Inhibit system from going down during crucial operations

joebonrichie:
https://www.freedesktop.org/wiki/Software/systemd/inhibit/

moss: prevents the system from going down during critical operations.
boulder: Allows the packager to keep suspend options enabled without the system going down during a long running build.

### DIFF
--- a/boulder/src/cli/build.rs
+++ b/boulder/src/cli/build.rs
@@ -11,6 +11,7 @@ use boulder::package::Packager;
 use boulder::{container, package, profile, timing, Env, Timing};
 use chrono::Local;
 use clap::Parser;
+use moss::signal::inhibit;
 use thiserror::Error;
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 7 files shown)


boulder/src/cli/build.rs
```diff
@@ -69,6 +69,27 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
...
+        Some("org.freedesktop.login1.Manager"),
+        "Inhibit",
+        &(
+            "shutdown:sleep:idle:handle-lid-switch",
+            "boulder",
+            format!("Build in-progress: {}", pkg_name),
+            "block",
+        ),
+    )?;
+    let _ = msg.body();
```
tarkah: It'd be good to inhibit during moss operations as well. Let's refactor this to a function in the moss crate, maybe under the `signal` module? `signal::inhibit_shutdown`?

moss/src/signal.rs
```diff
@@ -27,6 +28,25 @@ pub fn ignore(signals: impl IntoIterator<Item = Signal>) -> Result<Guard, Error>
     ))
 }
 
+// https://www.freedesktop.org/wiki/Software/systemd/inhibit/
+pub fn inhibit(what: Vec<&str>, who: String, why: String, mode: String) -> Result<message::Body, zbus::Error> {
```
tarkah: Do `what` and `mode` need to be exposed? Or will all uses of `inhibit` across the codebase use the same? Ideally we don't need to expose these to the rest of the codebase.

boulder/Cargo.toml
```diff
@@ -44,3 +44,4 @@ thiserror.workspace = true
 tokio.workspace = true
 url.workspace = true
 mailparse.workspace = true
+zbus.workspace = true
```
tarkah: You can remove this dep from `boulder`

moss/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tokio-util.workspace = true
 thiserror.workspace = true
 url.workspace = true
 xxhash-rust.workspace = true
+zbus.workspace = true
```
tarkah, line 41: This dep is missing in the `Cargo.toml` workspace deps

moss/src/signal.rs
tarkah: Return `Error` instead of `zbus::Error` so consumers only need to depend on this module.

PR #351, opened 2024-08-27
boulder: Update cargo_test macro as well

ReillyBrogan:


ikeycode (approved):

tarkah (commented):
Hmm not sure we should test with release profile by default, it'll be much slower to compile.

PR #354, opened 2024-09-04
Use fs-err crate for better filesystem-related error messages

jplatte:
For example before:
```text
Error: env: io: Permission denied (os error 13)
```
and after:
```text
Error: env: io: failed to create directory `/usr/share/boulder`: Permission denied (os error 13)
```


ikeycode (approved):
omg yes please

PR #356, opened 2024-09-17
boulder: Set SCHED_BATCH for builds

ReillyBrogan:
By default set the priority to SCHED_BATCH to hint to the scheduler that boulder is not interactive and to optimize scheduling for throughput. This should help improve system responsiveness during high CPU builds.

In case the user doesn't want this for some reason also add a flag to disable this behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,6 +247,7 @@ dependencies = [
  "stone_recipe",
  "strum",
  "thiserror",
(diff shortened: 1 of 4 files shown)


ermo (commented):
I see no reason this shouldn't go in.

ikeycode (approved):

PR #364, opened 2024-10-29
Feat/stone ffi

tarkah:


ikeycode (commented):
Mildly terrifying =) Was it not possible to autogenerate the header from the extern "C" functions..? My only nitpicks atm are missing docs/SPDX headers. I assume the renames are required for FFI exports, and unavoidably introduce the verbosity / far reach of changes?

tarkah:
> Was it not possible to autogenerate the header from the extern "C" functions..?

The header is being auto generated by cbindgen, checkout build.rs!

> I assume the renames are required for FFI exports, and unavoidably introduce the verbosity / far reach of changes?

Lol yeah cbindgen operates on type names only, not fully qualified paths, so any "duplicate" name, even from different modules, collide and it'll only emit one of them. So I had to switch the entire crate to fully qualified type syntax :/ 

> My only nitpicks atm are missing docs/SPDX headers.

👍 Yup we need 100% documentation and outlining safety & lifetimes will be important. Any necessary allocations happen on the rust side. The consumer needs to copy out any StoneStrings we allocate in Rust to C before destroying the reader and Rust drops it all. Since C is devoid of lifetimes, le documentation must be solid.

ermo:
Tentatively assigning this to alpha2.

libstone/cbindgen.toml
```diff
@@ -0,0 +1,24 @@
+include_guard = "STONE_H"
+language = "C"
+
```
joebonrichie, line 4:
```suggestion
cpp_compat = true.
```

For `appstream-generator` compatibility.

It's likely the generated header will change slightly

PR #367, opened 2024-11-21
Run boulder inside docker fixes

tarkah:
Fixes needed to run boulder inside docker:

- systemd is probably not pid 1 and we can't rely on inhibit working (don't make this a fatal error)
- We have some logic around term fg capturing to recapture ctrl-c functionality after invoking bash within user namespace. This fails when we don't have a controlling terminal. This can be ignored in those cases and shouldn't be fatal.


ermo:
@ikeycode Plz take a look at this ASAP. Thanks ^^'

ikeycode (approved):

PR #374, opened 2024-12-27
boulder: Add `%gtk_update_icon_cache` action macro

ermo:
Most icon themes benefit from having their icon cache pre-generated at build time.

The invocation for doing so follows a distinct pattern, hence this macro could perhaps be gainfully used as a simpler substitution.


ikeycode (approved):
le win

PR #375, opened 2024-12-27
boulder: Write a monitoring.yaml file with boulder new

joebonrichie:
Attempt to match the id from release-monitoring and any CPE from cpe-guesser.

### DIFF
--- a/boulder/src/draft.rs
+++ b/boulder/src/draft.rs
@@ -15,14 +15,21 @@ use url::Url;
 use crate::util;
 
 use self::metadata::Metadata;
+use self::monitoring::Monitoring;
 use self::upstream::Upstream;
 
 mod build;
 mod metadata;
+mod monitoring;
 mod upstream;
 
 pub struct Drafter {
     upstreams: Vec<Url>,
+}
+
+pub struct Draft {
+    pub stone: String,
+    pub monitoring: String,
 }
 
 impl Drafter {
@@ -30,7 +37,7 @@ impl Drafter {
         Self { upstreams }
     }
 
-    pub fn run(&self) -> Result<String, Error> {
+    pub fn run(&self) -> Result<Draft, Error> {
         // TODO: Use tempdir
         let extract_root = PathBuf::from("/tmp/boulder-new");
 
@@ -39,6 +46,9 @@ impl Drafter {
 
         // Build metadata from extracted upstreams
         let metadata = Metadata::new(extracted);
+
+        let monitoring = Monitoring::new(&metadata.source.name, &metadata.source.homepage);
+        let monitoring_result = monitoring.run()?;
 
         // Enumerate all extracted files
         let files = util::enumerate_files(&extract_root, |_| true)?
@@ -96,7 +106,10 @@ license     : UPDATE LICENSE
             metadata.upstreams(),
         );
 
-        Ok(template)
+        Ok(Draft {
+            stone: template,
+            monitoring: monitoring_result,
+        })
     }
 }
 
@@ -135,6 +148,8 @@ pub enum Error {
     AnalyzeBuildSystem(#[source] build::Error),
     #[error("upstream")]
     Upstream(#[from] upstream::Error),
+    #[error("monitoring")]
+    Monitoring(#[from] monitoring::Error),
     #[error("io")]
     Io(#[from] io::Error),
 }
--- a/boulder/src/draft/monitoring.rs
+++ b/boulder/src/draft/monitoring.rs
@@ -1,0 +1,221 @@
+// SPDX-FileCopyrightText: Copyright © 2020-2024 Serpent OS Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
+use std::io;
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::collections::HashMap;
+use thiserror::Error;
+use tui::Styled;
+
+#[derive(Serialize)]
+struct MonitoringTemplate {
+    releases: Releases,
+    security: Security,
+}
+
+#[derive(Serialize)]
+struct Releases {
+    id: Option<u32>,
+    rss: Option<String>,
+}
+
+#[derive(Serialize)]
+struct Security {
+    cpe: Vec<Cpe>,
+}
+
+#[derive(Serialize)]
+struct Cpe {
+    vendor: Option<String>,
+    product: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+struct Item {
+    id: u32,
+    name: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct Response {
+    items: Vec<Item>,
+    total_items: u32,
+}
+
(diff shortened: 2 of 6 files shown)


joebonrichie:
draft pending code cleanup

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,217 @@
...
+        let cpes = self.find_security_cpe(&self.name, &client)?;
+
+        self.write_monitoring(id, cpes)?;
+
+        Ok(())
+    }
+
+    fn create_reqwest_client(&self) -> reqwest::blocking::Client {
+        reqwest::blocking::Client::new()
+    }
```
tarkah: Nit: `self.create_reqwest_client()` can just be inlined (it's a single line function only used here). Also, it doesn't really make sense to have as a method of `Monitoring` since it doesn't actually need `&self` and could instead be a standalone function if we really needed it as a function.

boulder/src/draft.rs
```diff
@@ -39,6 +41,9 @@ impl Drafter {
         // Build metadata from extracted upstreams
         let metadata = Metadata::new(extracted);
 
+        let monitoring = Monitoring::new(metadata.source.name.clone());
+        Monitoring::run(&monitoring)?;
```
tarkah:
```suggestion
        let monitoring = Monitoring::new(metadata.source.name.clone());
        monitoring.run()?;
```

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn find_security_cpe(&self, name: &String, client: &reqwest::blocking::Client) -> Result<Option<Vec<Cpe>>, Error> {
+        let url = "https://cpe-guesser.cve-search.org/search";
+
+        let mut query = HashMap::new();
+        query.insert("query", [name]);
+
+        let resp = client.post(url).json(&query).send()?;
+
+        if !resp.status().is_success() {
+            println!("Failed to get monitoring CPE, error {}", resp.status());
```
tarkah: Might as well `return Ok(None);` here since it doesn't make sense to try and deserialize the response body if the request didn't succeed.

You can also do https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,217 @@
...
+            println!(
+                "{} | Find the correct ID for the project at https://release-monitoring.org/",
+                "Warning".yellow()
+            );
+            Ok(None)
+        }
+    }
+
+    fn find_security_cpe(&self, name: &String, client: &reqwest::blocking::Client) -> Result<Option<Vec<Cpe>>, Error> {
+        let url = "https://cpe-guesser.cve-search.org/search";
```
tarkah: You can make this a `const`

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fn find_monitoring_id(&self, name: &String, client: &reqwest::blocking::Client) -> Result<Option<u32>, Error> {
+        let url = format!("https://release-monitoring.org/api/v2/projects/?name={}", name);
+
+        let resp = client.get(&url).send()?;
+
+        if !resp.status().is_success() {
+            println!("Failed to get monitoring ID, error {}", resp.status());
+        }
```
tarkah: Same feedback here as below (`const` and `error_for_status`)

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn run(&self) -> Result<(), Error> {
+        let client = self.create_reqwest_client();
+
+        let id = self.find_monitoring_id(&self.name, &client)?;
+        let cpes = self.find_security_cpe(&self.name, &client)?;
+
+        self.write_monitoring(id, cpes)?;
+
+        Ok(())
+    }
```
tarkah: `run` doesn't describe this very well. Maybe `fetch_and_output`, but that name is a bit of a code smell. Does this API really need to be responsible over outputting the data?

Maybe we should just make this `Monitoring::fetch` which can be simplified to a module level function.. `monitoring::fetch(name) -> Monitoring` and then the caller can handle this type / write it however they want (`Monitoring` can be the name of what's currently `MonitoringTemplate`)

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,214 @@
...
+        if monitoring_template.security.cpe.unwrap_or_default().is_empty() {
+            let cpe_string = "cpe: []";
+            let cpe_marker = yaml_string.find(cpe_string).expect("security cpe marker not found");
+            yaml_string = yaml_string.replace(cpe_string, "cpe: ~");
+            let cpe_help_text = format!(
+                " # Last checked {}",
+                chrono::Local::now().date_naive().format("%Y-%m-%d")
+            );
+            yaml_string.insert_str(cpe_marker + cpe_string.len() - 1, &cpe_help_text);
+        }
```
tarkah: Is it even worth using `serde_yaml::to_string` vs building the formatted string ourselves since we're doing all these replaces?

```rs
let id = id.map_or_else(
 || "~ # https://release-monitoring.org/ and use the numeric id in the url of project".to_string(), 
  ToString::to_string
);

format!(r"---
releases:
  id: {id}
")
```

A bit more readable imo and more robust

joebonrichie: Potential multiple CPE matches make this more annoying

boulder/src/draft/monitoring.rs
```diff
@@ -0,0 +1,214 @@
...
+                " # https://release-monitoring.org/ and use the numeric id in the url of project";
+            yaml_string.insert_str(id_marker + id_string.len(), ID_HELP_TEXT);
+        }
+
+        if monitoring_template.releases.rss.is_none() {
+            let rss_string = "rss: null";
+            yaml_string = yaml_string.replace(rss_string, "rss: ~");
+        }
+
+        if monitoring_template.security.cpe.unwrap_or_default().is_empty() {
```
tarkah: `Option<Vec<_>>` is redundant since `Vec` can be empty. `cpe: Vec<_>` and then you can just check `monitoring_template.security.cpe.is_empty()` without `unwrap_or_default`

boulder/src/draft.rs
```diff
@@ -90,7 +95,7 @@ license     : UPDATE LICENSE
             metadata.upstreams(),
         );
 
-        Ok(template)
+        Ok((template, monitoring_result))
```
tarkah: Lets define a new struct and return it w/ named fields so the API is self describing and caller can't mix up the tuple values.

```rs
struct Draft {
  stone: String,
  monitoring:  String,
}

fn run() -> Result<Draft, Error>
```

tarkah (approved):
Awesome, great addition. Thanks @joebonrichie!

PR #379, opened 2024-12-29
Replace thiserror with SNAFU

jplatte:
If we decide to go this path, we should probably convert all of the crates.

This forces us to be much more explicit about error propagation and adding any relevant context, but makes that very convenient.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -641,8 +641,8 @@ version = "0.24.7"
 dependencies = [
  "fs-err",
  "nix 0.27.1",
+ "snafu",
  "strum",
- "thiserror 2.0.3",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


crates/config/src/lib.rs
```diff
@@ -74,13 +74,13 @@ impl Manager {
...
-        fs::create_dir_all(&dir).map_err(|io| SaveError::CreateDir(dir.clone(), io))?;
+        fs::create_dir_all(&dir).context(CreateDirSnafu)?;
 
         let path = dir.join(format!("{name}.{EXTENSION}"));
 
-        let serialized = serde_yaml::to_string(config)?;
+        let serialized = serde_yaml::to_string(config).context(SerializeSnafu)?;
 
-        fs::write(&path, serialized).map_err(|io| SaveError::Write(path, io))?;
+        fs::write(&path, serialized).context(WriteSnafu)?;
```
tarkah, line 83: What's the benefit of sanfu over thiserror? It seems like we're just exchanging map_err for a custom trait / .context but it provides the exact same utility in the end?

All the extra variants being added in this PR is great, but can just as easily be done with thiserror

jplatte: Two benefits:

- The `context` stuff is shorter, especially when adding extra context. This makes it practical to _not_ implement `From<UnderlyingError> for MyError` and in fact `snafu`'s derive macro never implements this.
- As a consequence, it becomes more apparent where errors are being converted, so it is much easier to notice / review whether a particular error variant is overused – I've seen some `Io` and `Nix` error variants, many of which would likely benefit from being split up.

tarkah: We can omit #[from] when defining variants with thiserror, but adding more friction here is definitely a good thing. The catch all Io, etc patterns are no good. If we can mostly solve that on the library portions of this repo then that'd be a big win, regardless of which error lib we use (I ultimately have no preference, they're effectively the same)

tarkah: I'm thankful you're looking at this, thank you! I'll jump back on matrix next week and it'd be good to connect on errors. It'd be great to focus on user facing errors next (moss / boulder CLI stuff) with anyhow / eyre. We're using thiserror there too and it's definitely the wrong approach.

tarkah: As an example I have this old draft:

https://github.com/serpent-os/tools/pull/170

PR #381, opened 2025-01-01
boulder: Numerous toolchain changes

ReillyBrogan:
Note: I bundled these together because most of them are pretty small and I worked on all of them at the same time, but I can split them out if needed (they're separate commits per-change).

Changes:
- Add build-id flags and enable it by default
- Default to compressing debug symbols with zstd. This reduces on-disk space of debug symbols significantly at the cost of higher package sizes.
- Define `-flto=%(jobs)` for GCC as "thin" LTO since it behaves roughly similar to the LLVM equivalent. Add `-flto=%(jobs) -flto-partition=one` as a new "full" variant of LTO for GCC.
- Add `-ffat-lto-objects` flags and enable them by default. These are needed if building static archives with LTO that are used by the alternate toolchain (IE if the static archives are built with GCC then fat-lto-objects is needed in order to link them with LLVM). Interestingly this also reduced binary sizes with LTO over not having the flag, likely because more information was preserved to the linker.
- Add some error flags recommended by Gentoo which can provide an early warning that LTO is likely to result in runtime issues. Enable by default.
- Enable thin LTO by default. Several other distros have switched to this as the default by now and it's better to deal with the fallout of it _now_ rather than when we have a few thousand extra packages.
- Add direct support for mold including automatically adding it as a builddep and setting the appropriate error flags
- Changed many dependencies added by macros to be `binary(*)` dependencies instead of named ones.

Testing:
- Did many, many builds between LLVM/GCC/Mold with the new flags in order to ensure that they worked everywhere as expected.

### DIFF
--- a/boulder/data/macros/arch/base.yaml
+++ b/boulder/data/macros/arch/base.yaml
@@ -101,6 +101,7 @@ defaultTuningGroups :
     - avxwidth
     - base
     - bindnow
+    - build-id
     - debug
     - fortify
     - frame-pointer
@@ -135,6 +136,11 @@ tuning              :
     - frame-pointer:
         enabled: no-omit-frame-pointer
         disabled: omit-frame-pointer
+
+    # Toggle build-id
+    - build-id:
+        enabled: build-id
+        disabled: build-id-none
 
     # Enable bindnow functionality
     - bindnow:
@@ -419,6 +425,16 @@ flags               :
         llvm:
             c         : "-fwhole-program-vtables"
             cxx       : "-fwhole-program-vtables"
+
+    # Enable build-id (ON)
+    - build-id:
+        ld        : "-Wl,--build-id=sha1"
+        rust      : "-C link-args=-Wl,--build-id=sha1"
+
+    # Disable build-id (OFF)
+    - build-id-none:
+        ld        : "-Wl,--build-id=none"
+        rust      : "-C link-args=-Wl,--build-id=none"
 
     # Enable ALL LLVM ICF optimisations (OFF)
     - icf-all:
--- a/boulder/data/macros/arch/base.yaml
+++ b/boulder/data/macros/arch/base.yaml
@@ -108,6 +108,7 @@ defaultTuningGroups :
     - frame-pointer
     - harden
     - icf
+    - lto
     - lto-errors
     - optimize
     - relr
@@ -197,7 +198,7 @@ tuning              :
                 enabled: lto-full
             - thin:
                 enabled: lto-thin
-        default: full
+        default: thin
 
     # Enable LTOextra. Requires the equivalent lto option
     - ltoextra:
@@ -206,7 +207,7 @@ tuning              :
                 enabled: ltoextra-full
             - thin:
                 enabled: ltoextra-thin
-        default: full
+        default: thin
 
     # Enable LTO errors
     - lto-errors:
@@ -413,7 +414,7 @@ flags               :
             d         : "-flto=full"
             ld        : "-flto=full"
 
-    # Enable Thin-LTO optimisations (OFF)
+    # Enable Thin-LTO optimisations (ON)
     - lto-thin:
         rust      : "-C lto=thin -C linker-plugin-lto -C embed-bitcode=yes"
         gnu:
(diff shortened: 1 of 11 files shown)


tarkah (approved):
LGTM

crates/stone_recipe/src/lib.rs
```diff
@@ -48,6 +48,8 @@ pub struct Recipe {
     pub tuning: Vec<KeyValue<Tuning>>,
     #[serde(default, deserialize_with = "stringy_bool")]
     pub emul32: bool,
+    #[serde(default, deserialize_with = "stringy_bool")]
+    pub mold: bool,
```
ikeycode, line 52: what happens when you want to support another experimental linker? Another root-level key? This part needs more consideration before changing here.

boulder/src/build/job/phase.rs
```diff
@@ -366,12 +372,18 @@ fn add_tuning(
...
             .filter_map(|flag| flag.get(tuning::CompilerFlag::D, toolchain)),
     );
-    let rustflags = fmt_flags(
+    let mut rustflags = fmt_flags(
         flags
             .iter()
             .filter_map(|flag| flag.get(tuning::CompilerFlag::Rust, toolchain)),
     );
 
+    if recipe.parsed.mold {
```
ikeycode, line 381: surely we can use substitutions for the ld plugin and not export full format strings here

boulder/data/macros/arch/base.yaml
```diff
@@ -420,6 +455,51 @@ flags               :
...
+    # Toggle LTO warning errors. If these throw an error it is likely that there will be runtime problems with LTO (ON)
+    - lto-errors:
+        gnu:
+            c         : "-Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
+            cxx       : "-Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
+        llvm:
+            c         : "-Werror=odr -Werror=strict-aliasing"
+            cxx       : "-Werror=odr -Werror=strict-aliasing"
+
+    # Enable build-id (ON)
```
ikeycode, line 477: im not really happy with this one - some times we'll encounter packages that ignore the LDFLAGS partly or fully, and sometimes we do it on purpose (for example bootstrap of toolchain) - so we need the build-id guarantee baked into the toolchain itself. We also rely on build-id for generation of debuginfo assets so I don't actually see a reason to support disabling this? Lastly our current default is actually `xxhash`, why the undocumented switch to sha1?

ikeycode (changes requested):

boulder/data/macros/arch/base.yaml
ReillyBrogan: Way ahead of you on baking it into the toolchain:
- LLVM: https://github.com/serpent-os/recipes/blob/llvm-cleanup/l/llvm/pkg/patches/config/0001-lld-Always-enable-build-id-and-use-20-byte-hashes.patch
- Mold: https://github.com/serpent-os/recipes/blob/blake3/m/mold/pkg/0001-Set-default-settings-for-compression-and-build.patch
- binutils: (Has the upstream default of `--build-id=sha1` already).

The purpose of allowing `--build-id=none` to be set is moreso a just-in-case thing. Since we're setting it at the toolchain level having `--build-id=sha1` in the flags is mostly so that packagers can see that it's there, and since it's baked in just removing the flag isn't going to disable it in the toolchain. We need an explicit `--build-id=none` to turn it off.

I don't really see it being necessary to disable, but some things like golang handle their own buildid generation and it's better to have the escape than not I think.

> Lastly our current default is actually xxhash, why the undocumented switch to sha1?

We discussed this in Matrix already but I'm going to open up an issue to track these flag changes.

boulder/src/build/job/phase.rs
ReillyBrogan: Not sure what you mean? We only need to add the flag for mold, not substitute an existing one.

ikeycode:
Per matrix convo toolchain was just a busted-ass notion and we'll clean up in port to KDL

ikeycode (approved):

PR #383, opened 2025-01-03
Enable a few allow-by-default clippy warnings

jplatte:
Apart from the first commit, the removal of commas inside formatting macros right before a closing parenthesis on the same line was also done manually (but I was too lazy to split it into its own commit).


ikeycode (approved):
thanks!!

PR #387, opened 2025-01-05
Use rustix for most parts of blitting

jplatte:
I can't benchmark this, and since blit is way faster than triggers once fs caches have warmed up, this likely isn't all that great of an improvement but hey.. I've done the work now, might as well make a PR.

### DIFF
--- a/moss/src/client/mod.rs
+++ b/moss/src/client/mod.rs
@@ -11,21 +11,22 @@
 use std::{
     borrow::Borrow,
     fmt, io,
-    os::{fd::RawFd, unix::fs::symlink},
+    os::{
+        fd::{AsFd as _, AsRawFd, BorrowedFd},
+        unix::fs::symlink,
+    },
     path::{Path, PathBuf},
     time::Duration,
 };
 
 use fs_err as fs;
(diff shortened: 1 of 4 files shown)


jplatte:
I don't think I've even run this a single time, please test before merging (or I'll do it in the next days). Should have made this a draft PR.

ikeycode:
(`EINVAL`)

jplatte:
I should really open my untested PRs as drafts 🫠

moss/src/client/mod.rs
```diff
@@ -712,39 +730,33 @@ impl Client {
...
                     // Regular file
                     _ => {
-                        linkat(
-                            Some(cache),
-                            fp.to_str().unwrap(),
-                            Some(parent),
-                            subpath,
-                            nix::unistd::LinkatFlags::NoSymlinkFollow,
-                        )?;
+                        rustix::fs::linkat(cache, fp, parent, subpath, AtFlags::SYMLINK_NOFOLLOW)?;
```
joebonrichie, line 742: EINVAL error can be resolved by using empty AtFlags instead of SYMLINK_NOFOLLOW

PR #389, opened 2025-01-05
Port root pivoting code from nix to rustix

jplatte:
Haven't tested it 😅 
But it compiles now and I got no more energy for this stuff for today.

### DIFF
--- a/crates/container/src/lib.rs
+++ b/crates/container/src/lib.rs
@@ -11,7 +11,6 @@ use std::sync::atomic::{AtomicI32, Ordering};
 
 use fs_err::{self as fs, PathExt as _};
 use nix::libc::SIGCHLD;
-use nix::mount::{mount, umount2, MntFlags, MsFlags};
 use nix::sched::{clone, CloneFlags};
 use nix::sys::prctl::set_pdeathsig;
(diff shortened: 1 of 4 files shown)


crates/container/src/lib.rs
```diff
@@ -247,23 +250,29 @@ fn pivot(root: &Path, binds: &[Bind]) -> Result<(), ContainerError> {
 
     let old_root = root.join(OLD_PATH);
 
-    add_mount(None, "/", None, MsFlags::MS_REC | MsFlags::MS_PRIVATE)?;
-    add_mount(Some(root), root, None, MsFlags::MS_BIND)?;
+    mount_change("/", MountPropagationFlags::REC | MountPropagationFlags::PRIVATE)
+        .map_err(|source| ContainerError::MountSetPrivate { source })?;
```
tarkah, line 254: Nit: if you use newtype enum pattern `MountSetPrivate(InnerError)` you can do

`.map_err(ContainerError::MountSetPrivate)`

jplatte: I know. I was following the `snafu` pattern already basically. Do you think I should change it for more consistency with what we already have, for now?

PR #393, opened 2025-01-06
moss: verify: query installdb directly

### LINKED ISSUE #392: moss: `state verify` fails if the current state is corrupt and moss no longer has access to the source .stone
**Context**

I had edited a few files belong to bash and starship, with each being installed from my local repo in my current moss state.

Before running the state verification command below, I had cleaned out my local repo.

```
(ldc-1.32.2)ermo@blueshift:~/repos/serpent-os/recipes/s/starship
$ sudo moss state verify
Verifying assets
Verifying states
Found 4 issues
 × Corrupt asset 1d05effada0b6afb89f668c10dea2772 - {"share/defaults/etc/bashrc"}
 × Corrupt asset 31ebb24fcc0242e52e5b980f8a677ea8 - {"share/defaults/etc/profile.d/40-prompt.sh"}
 × Corrupt asset 99915357e42cbbc01ea54e6d2f4e4f00 - {"share/defaults/etc/profile.d/50-starship.sh"}
 × Corrupt asset ac8229d3a3a960ab3cf4ed2e09bc5d6c - {"share/defaults/etc/profile"}
✔  Fixing issues, this will change your system state. Do you wish to continue?  · yes
Error: state: client: No metadata found for package Id("1cdc9f1adf19f68ffe2def3aee0862390449e69ab627303c676a668b1763c88e")
```

**Potential solution?**

In this specific error mode, would it perhaps make sense to:
- Offer to rollback to a known good state (which might or might not fix the corruption on its own)
- After the rollback, offer to re-run the state verify?

I believe that ^ would be a nice, automatic approach to encountering an unexpected (and unrecoverable?) state?

What other approaches might one fruitfully take here?

tarkah:
Issue packages can come from states that are not active and it's possible those packages comes from repos that are no longer available. Those packages still exist in our layout / install dbs until pruned from all states, but since we were querying via Registry API, it only considers packages from available repos or the active state.

Resolves #392

### DIFF
--- a/moss/src/client/verify.rs
+++ b/moss/src/client/verify.rs
@@ -16,7 +16,7 @@ use vfs::tree::BlitFile;
 
 use crate::{
     client::{self, cache},
-    package, runtime, signal, state, Client, Signal,
+    package, runtime, signal, state, Client, Package, Signal,
 };
 
 pub fn verify(client: &Client, yes: bool, verbose: bool) -> Result<(), client::Error> {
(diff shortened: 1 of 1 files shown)


ermo (commented):
@ikeycode I can confirm that this code ended up solving my issue, so if the logic is sound in your view, I think it's ready to go.

ikeycode (approved):
<3

PR #395, opened 2025-01-06
boulder: Normalize all python providers to lowercase

joebonrichie:
https://peps.python.org/pep-0503/#normalized-names

pip and python packaging APIs use case-insensitive names, until now this hasn't been a problem until we started providing the python "Jinja2" whilst dependencies were looking for "jinja2".

Apon landing in summit we need safety rebuilds of all python packages with this change.


ermo (approved):
LGTM, thanks!

tarkah (approved):
LGTM

A potential enhancement we might want to make when parsing stone files is to always lowercase these during decoding as well

ikeycode (approved):
very glad this is caught. ill get rebuilds prepped.

PR #396, opened 2025-01-07
boulder/draft: Default to autotools on unhandled build system

joebonrichie:
This will most likely be the closest match, plus we write out the recipe file unconditionally then much to deangst to our packagers.

### DIFF
--- a/boulder/src/draft.rs
+++ b/boulder/src/draft.rs
@@ -9,6 +9,7 @@ use fs_err as fs;
 use itertools::Itertools;
 use moss::Dependency;
 use thiserror::Error;
+use tui::Styled;
 use url::Url;
 
 use crate::util;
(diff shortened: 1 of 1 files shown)


boulder/src/draft.rs
```diff
@@ -54,9 +55,13 @@ impl Drafter {
...
-        let Some(build_system) = build.detected_system else {
-            return Err(Error::UnhandledBuildSystem);
-        };
+        let build_system = build.detected_system.unwrap_or_else(|| {
+            println!(
+                "{} | Unhandled build system! - Defaulting to autotools",
+                "Warning".yellow()
+            );
+            build::System::Autotools
+        });
```
tarkah, line 64: I see you're "logging" like this in a consistent format across various PRs.

We should really use `log` / some logging library like `env_logger` / `fern` / etc, and do this properly so they can be filtered out if need be. Then we can have some `moss::logging` module with some helpers to produce standardized log messages throughout the codebase

ikeycode:
For now I'll merge as its definitely necessary and I can't blame lack of logging on anyone but me :D

ikeycode (approved):

PR #399, opened 2025-01-08
boulder/draft: Fix github matching

joebonrichie:
We need to match against "special" providers first before falling back to the default simple case, as it'll break the match on first matcher.


ermo (approved):
LGTM, thanks!

tarkah (approved):

ikeycode (approved):

PR #402, opened 2025-01-09
boulder: Fix pattern matching

ReillyBrogan:
Boulder had a needless check for `starts_with` which was causing erroneous package splitting in several cases. For instance, consider the following simplified patterns:

```
- llvm-libs:
  paths:
    - /usr/lib/libLLVM.so.*

- llvm-devel:
  paths:
    - /usr/lib/libLLVM.so
```

With this setup the file `/usr/lib/libLLVM.so.18.1` should be patterned into `llvm-libs`, however since the `starts_with` match was checked first it would end up in `llvm-devel`. The point of this was to ensure that we could have "globs" that matched directories that matched subfiles/subdirectories and in order to preserve that behavior we instead match twice, once on the submitted pattern and again with `/**` appended to get a recursive glob.

Also while we're at it apply a minor adjustment to escape the directory so that all glob patterns work correctly.

### DIFF
--- a/boulder/src/package/collect.rs
+++ b/boulder/src/package/collect.rs
@@ -24,10 +24,20 @@ pub struct Rule {
 
 impl Rule {
     pub fn matches(&self, path: &str) -> bool {
-        self.pattern == path
-            || path.starts_with(&self.pattern)
-            || Pattern::new(&self.pattern)
-                .map(|pattern| pattern.matches(path))
+        if self.pattern == path {
+            return true;
+        }
+
+        // Escape the directory in case it contains characters that have special
+        // meaning in glob patterns (e.g., `[` or `]`).
+        let escaped_path = Pattern::escape(path);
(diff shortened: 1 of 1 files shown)


ReillyBrogan:
Found another bug, figuring it out

ikeycode:
are you sure..? check the manifest, the splitting appears fine, I had to fix some wild globs in it a while back

ermo:
@ikeycode From an engineering-led perspective, perhaps the smart thing to do here would be to set up a test suite for the pattern matching code, and having said test suite become the "specification" in a sense?

Would make the discussion very simple in that it'd simply be: "Well, what do the tests show?"

ikeycode:
Yeah we do need to extend tests

ikeycode:
Per matrix convo `toolchain` was just a busted-ass notion and we'll clean up in port to KDL

ikeycode (approved):

PR #407, opened 2025-01-15
moss: cli: Add blit stats

tarkah:
moss-blit-stats


ermo (approved):
Smart move on adding the extra stats re. symlinks/files/directories (for a potential future verbose stats mode I presume?)

PR #412, opened 2025-01-21
Dependency upgrades and other maintenance

jplatte:
I have not tested this (nor am I certain how to best do that). Much more straight-forward than my failed rustix PRs though.

I'm happy with rebasing this over time if keeping up with the ecosystem is not currently a priority. (I just like doing this sort of work and had fs-err `File::options` on my todo > got that merged today > I checked out where it would be useful > found `OpenOptions` usage in this repo > found more dependencies that were not at the latest version > found an unnecessarily large unsafe block while upgrading nix)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ dirs = "6.0.0"
 elf = "0.7.4"
 indicatif = "0.17.8"
 itertools = "0.14.0"
-fs-err = { version = "3.0.0", features = ["tokio"] }
+fs-err = { version = "3.1.0", features = ["tokio"] }
 futures-util = "0.3.31"
 glob = "0.3.1"
 hex = "0.4.3"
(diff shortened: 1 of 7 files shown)


ermo:
@jplatte Is this one rebase-able?

jplatte:
I forgot all about this! A bunch of the changes have already gone in, I think. Need to check about rustix, I tried to upgrade it again yesterday (having forgotten about this) but hit a small roadblock. Now I can check how past me solved it! 😅

jplatte:
I kicked out the nix upgrade because I was uncertain about it. Can be a separate PR. This is ready for review :)

ermo:
@tarkah I think we will want this to go into the upcoming alpha1 os-tools release (to be used when we do the full repo rebuild), so if you could find the time to absorb and review this PR, that would be genuinely _mega_!

jplatte:
It shouldn't really matter when this is merged, there are no functional changes.

ermo (approved):
LGTM, thanks!

PR #413, opened 2025-01-22
boulder/macros: Add ccache and Qt macros

ReillyBrogan:
- Add simple ccache stats macros since I use them frequently
- Add some basic patterns for Qt5/Qt6
- Add Qt5 and Qt6 macros


ikeycode (approved):
lgtm

PR #421, opened 2025-02-10
moss: parallelize 'state verify'

### LINKED ISSUE #390: moss: Explore making `moss state verify` able to use multiple threads when verifying
Currently, it looks like moss is stuck verifying on a single core at a time.

Perhaps it would be fruitful in a performance sense to parallelize this feature?

joebonrichie:
Use rayon to iterate over entries in parallel and use a Arc<Rwlock<>> to update the vec of issues atomically.

Resolves #390.


ermo:
@joebonrichie Please rebase this -- it fails with `just get-started` :smiling_face_with_tear:

ermo:
@joebonrichie :ping_pong:

PR #423, opened 2025-02-13
add safeguard to prevent moss from eating your current operating system

cameronbraid:
Only provide a default value for "root" global argument if the current system is identified as a serpent os.

Currently uses the MACHTYPE env for detection

Manually tested on a arch system :

```
> moss list installed
Error: args: Root directory is required, but not provided
```

And when we fake then env  we get an expected error

```
> MACHTYPE=serpent moss list installed
Error: list: client: db: diesel connection: Unable to open the database file
```

### DIFF
--- a/moss/src/cli/mod.rs
+++ b/moss/src/cli/mod.rs
@@ -4,7 +4,10 @@
 
 use std::{env, fs, io, path::Path, path::PathBuf};
 
-use clap::{Arg, ArgAction, Command};
+use clap::{
+    builder::{IntoResettable, OsStr},
+    Arg, ArgAction, Command,
+};
 use clap_complete::{
     generate_to,
     shells::{Bash, Fish, Zsh},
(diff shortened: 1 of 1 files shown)


Quackdoc:
Would it not be better to check `/usr/lib/os-release` ? I know MACHTYPE can sometimes change depending on shell.

also this seems like it would limit moss to working on serpent as the system package manager. perhaps there could be a better way of doing this. Could this be set via config or something?

cameronbraid:
What about changing the default back to `/` and add an assertion that "$root/.moss" folder is present

cameronbraid:
Although, this might suggest that a separate `moss init` command should be added as the only means to create a .moss installation... then all other commands could fail if its not a valid installation

ermo:
The idea of having a well thought out `moss init` feature sounds compelling.

This migh also tie nicely into potentially setting up the proper minimum SELinux policy for /.moss and /usr in the future.

ermo:
@cameronbraid You still up for working on this?

PR #429, opened 2025-02-21
boulder/draft: Use infer to get extension type to perform extraction

joebonrichie:
Using infer will help us catch specalized extension types

e.g.

`.whl` is a `.zip`
`.gem` is a `.tar`


ikeycode:
smart approach i like it =)

ikeycode (approved):

PR #438, opened 2025-03-05
Arch: aarch fixes and add riscv64 definition 

Vixea:
requires a unreleased version of llvm(already in a released version of gcc) for the unused RISC-V part


Vixea:
Imma open a new one, noticed a few mistakes and I don't like how aarch was autoformated

PR #448, opened 2025-03-31
Arch: aarch fixes and add riscv64 definition

Vixea:
RISC-V:
    LLVM: Doesn't yet support generic-ooo
AARCH64:
    GCC: While passing -m64 in clang is fine, its not for gcc


ReillyBrogan:
As far as I can tell that riscv flag missed the llvm 20 release and will be in llvm 21. Is there an alternate flag we can use instead?

Vixea:
I'm just going to add the patch to llvm 20

Vixea:
Yea there's others but not very future proof

ermo (approved):
LGTM, thanks!

PR #450, opened 2025-04-03
boulder/macros: Update build triple

ReillyBrogan:


boulder/data/profile.d/default-x86_64.yaml
```diff
@@ -1,6 +1,6 @@
 default-x86_64:
   repositories:
     volatile:
-      uri: "https://packages.serpentos.com/volatile/x86_64/stone.index"
-      description: "Volatile Serpent OS repo"
+      uri: "https://packages.aerynos.com/volatile/x86_64/stone.index"
+      description: "Volatile Aeryn OS repo"
```
ermo, line 5: This needs to be AerynOS (sorry!).

ermo (changes requested):
Small nit, otherwise good to go.

boulder/data/profile.d/default-x86_64.yaml
ermo: I'll do this after I land it to main, as I need to rebase my own os-tools branch on this anyway.

ermo (approved):
Let's get this in.

PR #451, opened 2025-04-03
boulder/macros: Update cmake/ninja/meson macros

ReillyBrogan:
Cmake:
- Re-introduce support for using make as the build server
- Switch to invoking the build server via cmake directly
- Update ctest options
- Fix handling of build flags
  - Build flags are now imported from environment directly
  - `-DNDEBUG` is now correctly added to build flags when using `Release` profile
- Enable verbose makefiles and disable stripping by default

Ninja:
- Add ninja-specific macros for building, installing, and testing

Meson:
- Remove unnecessary exports of build flags
- Update build options


ermo (approved):
LGTM, thanks!

PR #453, opened 2025-04-05
boulder: Normalize python providers

cybin:
Boulder generates providers by using the module's name. If module's name contains a '_', it will cause a provider/dependency mismatch. This is solved by simply adding additional providers by replacing '_' with '-' in the module's name.

I tested it by rebuilding python-typing_extensions and it got resolved, even if the module is looking for python(typing-extensions(3.11)), which got generated by boulder too.


ermo (changes requested):
Please address clippy lints.

ermo:
@joebonrichie Do you have anything to add to this before I merge?

ermo (approved):
Let's try this on for size.

PR #454, opened 2025-04-06
moss/request: Add proxy variable support

Elecleus:


moss/src/request.rs
```diff
@@ -20,13 +20,70 @@ use crate::environment;
...
+            .to_str()
+            .unwrap(),
+        )
+        .unwrap(),
+    )
+}
+
+fn get_http_proxy() -> Option<reqwest::Proxy> {
+    let lower_case = var_os("http_proxy");
+    let upper_case = var_os("HTTP_PROXY");
```
tarkah: We can check all 4 vars in order of priority and take the first one and simplify all this to a single check:

`["HTTPS_PROXY", ..].iter().find_map(var_os)`

PR #464, opened 2025-05-06
Upgrade some dependencies

jplatte:
Not all of them though. Can upgrade more in follow-up PRs. This is to get the repo into a state where we can have as few duplicate dependencies in `infra` as reasonable without much work :)


ermo (approved):
LGTM, thanks!

PR #469, opened 2025-05-07
boulder: Fix machine_isa for alt isa

Vixea:


ermo:
@jplatte Is this good to go do you think?

jplatte:
A few thoughts:

- this is definitely an improvement
- why do we `unwrap_or_default` (pre-existing)? is it ever correct to work with an empty string?
- in the upstream bug report, there was talk of an `em_` prefix, this strips `EM_`; do we need to also look for the lowercase variant?

Vixea:
We do not need to worry about a lowercase version either to_lowercase handles that or strip_prefix is case insensitive

jplatte:
I am very certain that `strip_prefix` is *not* case insensitive. Where did you get that from?
Also, `to_lowercase` is only called afterwards.

Vixea:
Ahh yes, anyway the answer is still no unless the library author changes that as every instance of em_ is capitalized only RISC-V doesn't have the em apparently

jplatte (approved):
Okay, based on the understanding that `e_machine_to_str` returns an all-uppercase string, this looks good. Maybe update your upstream bug report from talking about `em_` to talking about `EM_` though, to avoid confusion :)

ermo (approved):
Approved by @jplatte. Thanks!

PR #471, opened 2025-05-08
boulder: Remove assembler directive

ReillyBrogan:
This was causing build failures due to causing compiler failures with `-Werror`


ermo (approved):
I hit these myself during mass rebuilds, so this is very much appreciated in terms of ergonomics. Thanks!

PR #473, opened 2025-05-08
boulder: Add CCACHE_BASEDIR to improve cache hit rates

ReillyBrogan:
ccache resolves absolute paths to source files and uses that path as part of the cache key. In practice this means that changing the build directory invalidates all builds using a previous build directory, and given that the source archive name typically ends up being part of the build path (for instance `/mason/build/x86_64/llvm-project-20.1.4.src.tar.xz/`) the end result is that updating the version of a package is enough to invalidate all ccache entries from a previous version.

To fix this we can use the CCACHE_BASEDIR environmental variable. If set ccache will strip the value of it from the cache key (for instance `/mason/build/x86_64/llvm-project-20.1.4.src.tar.xz/src/foo.c` with a `CCACHE_BASEDIR=/mason/build/x86_64/llvm-project-20.1.4.src.tar.xz/` will be stored as `src/foo.c`) which means that ccache entries can persist between version updates. Given that ccache already hashes the file contents as part of the cache key this shouldn't cause any issues and should improve cache hit rates quite a bit.

Also while we're at it add macro definitions for the zsh, fish, and elvish shell completions directories since that's a trivial change not really deserving of a separate PR.

### DIFF
--- a/boulder/data/macros/arch/base.yaml
+++ b/boulder/data/macros/arch/base.yaml
@@ -3,28 +3,31 @@
 definitions:
 
     # Basic variables required for packages to build correctly
-    - libsuffix      : ""
-    - prefix         : "/usr"
-    - bindir         : "%(prefix)/bin"
-    - sbindir        : "%(prefix)/sbin"
-    - includedir     : "%(prefix)/include"
-    - datadir        : "%(prefix)/share"
-    - localedir      : "%(datadir)/locale"
-    - infodir        : "%(datadir)/info"
-    - mandir         : "%(datadir)/man"
-    - docdir         : "%(datadir)/doc"
-    - vendordir      : "%(datadir)/defaults"
-    - completionsdir : "%(datadir)/bash-completion/completions"
-    - tmpfilesdir    : "%(prefix)/lib/tmpfiles.d"
-    - sysusersdir    : "%(prefix)/lib/sysusers.d"
-    - udevrulesdir   : "%(prefix)/lib/udev/rules.d"
-    - localstatedir  : "/var"
-    - sharedstatedir : "%(localstatedir)/lib"
-    - runstatedir    : "/run"
-    - sysconfdir     : "/etc"
-    - libdir         : "%(prefix)/lib%(libsuffix)"
-    - libexecdir     : "%(libdir)/%(name)"
-    - builddir       : "serpent_builddir"
+    - libsuffix          : ""
+    - prefix             : "/usr"
+    - bindir             : "%(prefix)/bin"
+    - sbindir            : "%(prefix)/sbin"
+    - includedir         : "%(prefix)/include"
+    - datadir            : "%(prefix)/share"
+    - localedir          : "%(datadir)/locale"
+    - infodir            : "%(datadir)/info"
+    - mandir             : "%(datadir)/man"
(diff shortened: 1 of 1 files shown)


boulder/data/macros/arch/base.yaml
```diff
@@ -3,28 +3,31 @@
...
+    - bindir             : "%(prefix)/bin"
+    - sbindir            : "%(prefix)/sbin"
+    - includedir         : "%(prefix)/include"
+    - datadir            : "%(prefix)/share"
+    - localedir          : "%(datadir)/locale"
+    - infodir            : "%(datadir)/info"
+    - mandir             : "%(datadir)/man"
+    - docdir             : "%(datadir)/doc"
+    - vendordir          : "%(datadir)/defaults"
+    - completionsdir     : "%(datadir)/bash-completion/completions"
```
ermo, line 17: How about making this `%(bashcompletionsdir)` and then doing a backwards compat / default definition of `%(completionsdir)`  defined as  `%(bashcompletionsdir)` to make the pattern generic?

boulder/data/macros/arch/base.yaml
```diff
@@ -3,28 +3,31 @@
...
+    - zshcompletionsdir  : "%(datadir)/zsh/site-functions"
+    - fishcompletionsdir : "%(datadir)/fish/vendor_completions.d"
+    - elvishcompletions  : "%(datadir)/elvish/lib"
```
ermo, line 20: Does this macro omit the customary dir suffix for a specific reason...?

i.e. `%(elvishcompletionsdir)` vs. the current `%(elvishcompletions)`

ermo (changes requested):
Great cover letter and good initiative (with a couple of small nits for the shell completions bit)

boulder/data/macros/arch/base.yaml
ReillyBrogan: Yes, the specific reason is that that's the directory that elvish uses for vendor completions :shrug:.

boulder/data/macros/arch/base.yaml
ReillyBrogan: I'd rather just do a find/replace across the entire repo to switch to the new definition, but there's too much in flight right now to really do that.

boulder/data/macros/arch/base.yaml
ermo: But it's still a directory, right...?

(I've updated my original question to give it the proper context. Sorry that I didn't think to do that originally)

ermo:
Just a heads up that I am thinking that I will probably do switch to a format that looks like `%($shell_completionsdir)` to make it easier to read and to make the shell part stand out.

Once we get that in, I'll take care of doing a companion PR for the recipes that still use plain %(completionsdir) for bash completions.

boulder/data/macros/arch/base.yaml
ermo: Going to fix this after landing.

ermo (approved):
Will fix up remaining nits after merging.

ermo:
Cannot be merged because I made a fat finger mistake.

Closing in favour of https://github.com/AerynOS/os-tools/pull/483 and https://github.com/AerynOS/os-tools/pull/484

PR #475, opened 2025-05-09
boulder: Add support for compressman option

joebonrichie:
Compresses any man/info pages with zstd. A tradeoff between slightly bigger .stone files vs smaller installed size.

Tested
- nano (contains man and info files)
- xz (contains lots of man symlinks)


ermo:
@jplatte Is this perhaps something you would be interested in doing an "idiomatic rust" review of for mentoring and quality assurance purposes?

boulder/src/package/analysis.rs
```diff
@@ -51,7 +52,7 @@ impl<'a> Chain<'a> {
     }
 
     pub fn process(&mut self, paths: impl IntoIterator<Item = PathInfo>) -> Result<(), BoxError> {
-        println!("│Analyzing artefacts (» = Include, × = Ignore)");
+        println!("│Analyzing artefacts (» = Include, × = Ignore, ^ = Replace)");
```
ermo, line 55: NGL, I love this addition.

boulder/src/package/analysis/handler.rs
```diff
@@ -1,5 +1,10 @@
+use filetime::FileTime;
+use itertools::Itertools;
+use std::fs::File;
+use std::io::{BufReader, BufWriter, Write};
 use std::{
```
jplatte: Could you merge all of these `std` imports into one `use` statement?

There's a nightly rustfmt option for it as well. I'll check whether we configure it in any of the other repos.

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,88 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
 
     Ok(Decision::NextHandler.into())
 }
+
+pub fn compressman(bucket: &mut BucketMut, info: &mut PathInfo) -> Result<Response, BoxError> {
```
jplatte: It may be slightly controversial, but I side with clippy here.

```suggestion
pub fn compressman(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response, BoxError> {
```

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,88 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
...
+    /* If we have a man/info symlink update the link to the compressed file */
+    if info.path.is_symlink() {
+        let new_original = format!("{}.zst", fs::canonicalize(&info.path)?.display());
+        let new_link = format!("{}.zst", &info.path.display());
+
+        /*
+         * Depending on the order the files get analysed the new compressed file may not yet exist,
+         * compress it _now_ so the correct metadata src info is returned to the binary writer.
+         */
+        if !std::path::Path::new(&new_original).exists() {
```
jplatte: I think new code should generally prefer to use `std::fs::exists`, because `Path.exists()` turns any errors into `false` instead of reporting them.

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,88 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
...
+        return Ok(Decision::ReplaceFile {
+            newpath: generated_path,
+        }
+        .into());
+    }
+
+    let mut compressed_file = PathBuf::from(format!("{}.zst", info.path.display()));
+
+    /* We may have already compressed the file if we encountered a symlink to this file first */
+    if !&compressed_file.exists() {
```
jplatte: Same here, can also use `.try_exists()?` if you need the path after, as seems to be the case here.

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,88 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
...
+    let mut generated_path = PathBuf::new();
+
+    let metadata = fs::metadata(&info.path)?;
+    let atime = metadata.accessed()?;
+    let mtime = metadata.modified()?;
+
+    /* If we have a man/info symlink update the link to the compressed file */
+    if info.path.is_symlink() {
+        let new_original = format!("{}.zst", fs::canonicalize(&info.path)?.display());
+        let new_link = format!("{}.zst", &info.path.display());
```
jplatte, line 249: `Path::display` is only made for display purposes, I believe it turns non-utf8 into � (the replacement character). I'd recommend we add a small helper `add_extension` that's based on `Path::extension` + `Path::set_extension` (there's an unstable `Path::add_extension` that we can switch to once it becomes stable).

ermo:
@joebonrichie :ping_pong:

ermo (changes requested):
Changes requested per Jonas' review.

boulder/src/package/analysis/handler.rs
joebonrichie: For `new_link` the path doesn't yet exist before it's symlinked so the `format!()` usage is retained there.

jplatte: What does the path not existing have to do with this?

ermo: > What does the path not existing have to do with this?

The way I am reading it, the point Joey might be making is that you cannot canonicalize a path that doesn't yet exist on the FS?

@joebonrichie is ^ the correct way of looking at it...?

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,97 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
...
+
+    let metadata = fs::metadata(&info.path)?;
+    let atime = metadata.accessed()?;
+    let mtime = metadata.modified()?;
+
+    // TODO: Replace usage with .with_added_extension() when it becomes stable #127292
+    fn add_extension(path: &mut PathBuf, extension: &str) -> PathBuf {
+        path.set_extension(extension);
+        path.to_path_buf()
+    }
```
jplatte: Okay, so this isn't what I meant. I looked more closely at the docs, and it seems like really the best way to add an extension (as long as `with_added_extension` is unstable) is to go via `file_name`.. The current implementation is actually equivalent to `with_extension` which *overwrites* the (last piece of) the file extension, which I bet is why you were getting a canonicalization error.

I'm pretty sure this does the thing we want:

```suggestion
    // TODO: Replace usage with .with_added_extension() when it becomes stable #127292
    fn with_added_extension(path: &Path, extension: &str) -> PathBuf {
        match path.file_name() {
            Some(file_name) => {
                let mut file_name = file_name.to_owned();
                file_name.push(extension);
                path.with_file_name(file_name)
            }
            None => path.to_owned(),
        }
    }
```

boulder/src/package/analysis/handler.rs
```diff
@@ -203,3 +208,97 @@ pub fn cmake(bucket: &mut BucketMut<'_>, info: &mut PathInfo) -> Result<Response
...
+        // FIXME: replace with add_extension()
+        //        failed to canonicalize path {symlink} No such file or directory
+        let new_link = format!("{}.zst", &info.path.display());
+        //let new_link = add_extension(&mut info.path, "zst");
+
+        /*
+         * Depending on the order the files get analysed the new compressed file may not yet exist,
+         * compress it _now_ so the correct metadata src info is returned to the binary writer.
+         */
+        if !std::fs::exists(&new_original)? {
```
jplatte, line 256: Looks like we need to bump the `rust-version` in `Cargo.toml` to 1.81.0 for this. @ermo is that a thing we can do just like that, or do we need to make sure boulder still builds with 1.81.0? (due to what AerynOS packages right now or w/e)

ReillyBrogan: Rust 1.81 is a reasonable increase in MSRV. I think we decided on CURRENT_RUST minus three versions as our maximum MSRV at any time, which would technically allow us to go up to 1.83 (though I'm not particularly fond of increasing the MSRV above what we actually need).

AerynOS currently has 1.85.1. I haven't updated to 1.86.0 due to all the toolchain PRs in flight and we'll be going straight to 1.87 later this month.

ermo: The Cargo MSRV has now been bumped to 1.81 as part of vixea's https://github.com/AerynOS/os-tools/pull/480 PR that I just merged.

ermo (approved):
I am going to take it from here with Jonas' help, as this is now a blocker that we need to resolve ASAP.

@joebonrichie Thank you for taking the time to rebase this and for your efforts in getting it ready!

Issue #476, opened 2025-05-09
boulder: Add user-customisable buildroot package install set to improve packager UX


ermo:
The motivation for this is that we want the roots that get installed on package build to be as small as possible.

The conflicting motivation is that we also want to offer a good packaging experience when troubleshooting, which also includes having a decent amount of packages in the boulder chroot when using packaging `%break_continue` or `%break_exit` breakpoints to be dropped to a chroot shell.

This issue proposes two improvements whenever boulder spots a breakpoint or is invoked with the chroot argument:

- [ ] Add the user `~/.gitconfig` to the root of the build root. This will make it easier to do the `git init && git add . && git commit -m "Initial commit"` dance and have the author and e-mail be pulled from `~/.gitconfig`.
- [ ] Add the ability to specify a list of additions to the boulder chroot package set, to be installed whenever a recipe breakpoint is added. I am unsure if we can make this work well with an after-the-fact boulder chroot command, so it is possible that we may have to accept that this additional user-controlled set of packages is _always_ installed if the user package set config file is present.

PR #477, opened 2025-05-09
boulder: use /usr/bin/bash for script shebangs

ermo:
With this change, the build scripts generated by boulder from the stone.yaml recipes are no longer affected by the /usr/bin/sh symlink, which means that we can now safely point that to dash for a potential increase of up to +20% faster configure runs for e.g. gettext.

This commit also creates dedicated *_with_bash fallback autotools actions which should ensure that stubborn packages whose build systems rely on the presence of a bash shell can now simply use the *_with_bash actions.

In addition, add helix, less and ripgrep to roots for a nicer chroot experience when troubleshooting or writing/testing patches inside the chroot (yes, this is a really nice workflow actually).

TODO: The extra chroot packages really ought to be user-controlled (with a decent, if minimal fallback) as a boulder config option, along with optionally copying in the current user's .gitconfig to the root for easier patching inside chroots. See https://github.com/AerynOS/os-tools/issues/476 for the details.

### DIFF
--- a/boulder/data/macros/actions/autotools.yaml
+++ b/boulder/data/macros/actions/autotools.yaml
@@ -1,49 +1,112 @@
 actions:
 
-    - configure:
-        description: Perform ./configure with the default options
-        command: |
-            test -x ./configure || ( echo "%%configure: The ./configure script could not be found" ; exit 1 )
-            ./configure %(options_configure)
-        dependencies:
-            - binary(autoconf)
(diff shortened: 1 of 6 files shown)


ermo:
@ReillyBrogan I'm going cross-eyed looking at these macros. Would you mind giving them a quick once over for sanity before I merge this?

PR #487, opened 2025-05-16
boulder: Fix Qt/KDE macros

ReillyBrogan:
Because we're constructing the cmake command manually (as opposed to say adding specific qt cmake options) we need to manually add binary(ninja) to the cmake deps.


ermo (approved):
Looks like we could use this for rebuilds.

Issue #231, opened 2024-05-03
boulder: Switch existing emul32 packages to `386` ISA


ReillyBrogan:
This is a tracking issue for the fact that the rust boulder uses `386` for the ISA where the dlang boulder used `x86`. To resolve this (per discussion in Matrix), we need to do the following:

- [x] Add duplicate provides to all emul32 packages to provide both `386` and `x86` for all provides fields. Override the depends to use `x86` INSTEAD OF `386` - serpent-os/moss#223
- [x] Rebuild all emul32 packages with that boulder
- [x] Change the depends field to `386` instead of `x86`
- [x] Rebuild all emul32 packages with that boulder
- [x] Strip out the compat code so that emul32 package only provide and depend on `386`
- [ ] (Optional) Rebuild emul32 packages again. This is optional because it will happen naturally over time.

ermo:
We should take a look at whether we can utilise this approach to "wash out" the dlang boulder .stones when the time comes.

I am unclear on the consequences on the dlang side in terms of summit / libmoss.

PR #493, opened 2025-05-28
boulder: Emul32 migration part 3

ReillyBrogan:
For https://github.com/AerynOS/os-tools/issues/231


ermo (approved):
Let the great flush commence!

PR #502, opened 2025-06-10
boulder: Fix phase timing output once and for all

ermo:
Add tests to demonstrate that the phase timing output is fixed.


boulder/src/timing.rs
```diff
@@ -279,3 +279,68 @@ fn fmt_progress(elapsed: Duration, total: Duration) -> String {
...
+    fn test_minutes_and_seconds2() {
+        let elapsed = Duration::from_secs(61);
+        assert_eq!(fmt_elapsed(elapsed), "     1m01.00s");
+    }
+
+    #[test]
+    fn test_hours_minutes_and_seconds() {
+        let elapsed = Duration::from_secs(3600);
+        assert_eq!(fmt_elapsed(elapsed), "  1h    0.00s");
+    }
```
tarkah, line 309: You may enjoy the `test_case` crate to define all these test cases a bit easier

jplatte (approved):

boulder/src/timing.rs
ermo: Will probably do this in a follow-up PR. Thanks for the tip!

PR #504, opened 2025-06-13
boulder: Add `%cargo_vendor` action macro

HigherOrderLogic:
This enables packagers to lock dependencies for Rust-based packages.

### DIFF
--- a/boulder/data/macros/actions/cargo.yaml
+++ b/boulder/data/macros/actions/cargo.yaml
@@ -16,6 +16,14 @@ actions:
         dependencies:
             - binary(cargo)
 
+    - cargo_vendor:
+        description: Vendor all dependencies
+        command: |
+            mkdir -pv ./.cargo
+            cargo vendor -v --locked >> ./.cargo/config.toml
+        dependencies:
+            - binary(cargo)
+
     - cargo_build:
         description: Build the rust project
         command: |


HigherOrderLogic:
> If you update your actual commit to look the same, it looks nicer in our git history.


Should I just squash the commits and force push the branch?

boulder/data/macros/actions/cargo.yaml
```diff
@@ -19,7 +19,8 @@ actions:
     - cargo_vendor:
         description: Vendor all dependencies
         command: |
-            cargo vendor -v --locked
+            [ -d ./.cargo ] || mkdir .cargo
```
ermo: Please use bash syntax `[[ ]]` (the macros will explicitly be executed by bash, and `[` is actually a coreutils binary).

EDIT: Actually, why not just do a `mkdir -pv ./.cargo`?

ermo (changes requested):
Just a small nit.

boulder/data/macros/actions/cargo.yaml
HigherOrderLogic: > Actually, why not just do a `mkdir -pv ./.cargo`?

Good idea, I have never use that option so I dont know about it.

ermo (approved):
LGTM, thanks!

And welcome to AerynOS development ^^'

PR #514, opened 2025-06-17
moss: fix tx provider resolution order for sync

tarkah:
`sync` needs to consider `available` packages _before_ `installed` packages (we're updating things!) so we don't keep around stale packages that have updates.

Previously this was captured by the "pinned" concept but this was just a mechanism to ensure the "available" update was chosen before something "installed" due to how we prioritized installed provider lookup before available. Because we now reorder this and available happens before installed, "pinned" isn't needed anymore.

I tested the old "pin" behavior w/ resolving available over installed vs the new approach and the "sync" results were identical, but this was only one test case. The "install" and "remove" behavior is unchanged.

### DIFF
--- a/moss/src/registry/transaction.rs
+++ b/moss/src/registry/transaction.rs
@@ -5,7 +5,7 @@
 use dag::Dag;
 use thiserror::Error;
 
-use crate::{package, Provider, Registry};
+use crate::{Provider, Registry, package};
 
 enum ProviderFilter {
     /// Must be installed
--- a/moss/src/cli/remove.rs
+++ b/moss/src/cli/remove.rs
@@ -2,23 +2,23 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use clap::{arg, ArgMatches, Command};
+use clap::{ArgMatches, Command, arg};
 use itertools::{Either, Itertools};
 use std::collections::BTreeSet;
 use thiserror::Error;
 
 use moss::{
+    Installation, Provider,
     client::{self, Client},
     environment,
     package::Flags,
     registry::transaction,
     state::Selection,
-    Installation, Provider,
 };
 use tui::{
-    dialoguer::{theme::ColorfulTheme, Confirm},
-    pretty::autoprint_columns,
     Styled,
+    dialoguer::{Confirm, theme::ColorfulTheme},
+    pretty::autoprint_columns,
 };
 
 pub fn command() -> Command {


moss/src/cli/sync.rs
```diff
@@ -168,12 +168,14 @@ pub fn handle(args: &ArgMatches, installation: Installation, debug: bool) -> Res
...
+    // For each explicit package, replace it w/ it's sync'd change (if available)
     // or return the original package
     let with_sync = packages
         .iter()
-        .map(|p| {
-            let is_explicit = p.flags.explicit;
+        .filter_map(|p| {
+            if !p.flags.explicit {
+                return None;
+            }
```
jplatte, line 178: I think a separate `filter(!explicit)` + `map` would be a little clearer here (I read over the remaining closure body to look for `?`s or `None` returns that could also result in filtering out here).

moss/src/registry/transaction.rs
```diff
@@ -37,43 +39,27 @@ pub struct Transaction<'a> {
     // unique set of package ids
     packages: Dag<package::Id>,
 
-    /// packages which are always considered first
-    /// during [`ProviderFilter::Pinned`] but
-    /// aren't part of `packages` DAG
-    pinned_providers: Vec<package::Id>,
+    // Dependency lookup strategy
```
jplatte:
```suggestion
    /// Dependency lookup strategy
```

moss/src/registry/transaction.rs
```diff
@@ -95,23 +81,23 @@ impl Transaction<'_> {
     }
 
     /// Update internal package graph with all incoming packages & their deps
-    #[tracing::instrument(skip_all, fields(?lookup))]
-    fn update(&mut self, incoming: Vec<package::Id>, lookup: Lookup) -> Result<(), Error> {
+    #[tracing::instrument(skip_all, fields(lookup = %self.lookup))]
```
jplatte, line 79:
```suggestion
    #[tracing::instrument(skip_all, fields(lookup = ?self.lookup))]
```

moss/src/registry/transaction.rs
```diff
@@ -14,18 +14,20 @@ enum ProviderFilter {
...
-    // Look beyond installed/selections
-    All(Provider),
+    // Available in upstream repositories
+    Available(Provider),
 }
 
-#[derive(Clone, Copy, Debug)]
-enum Lookup {
+/// Dependency lookup strategy
+#[derive(Clone, Copy, Debug, strum::Display)]
```
jplatte, line 22:
```suggestion
#[derive(Clone, Copy, Debug)]
```

moss/src/registry/transaction.rs
```diff
@@ -95,23 +81,23 @@ impl Transaction<'_> {
     }
 
     /// Update internal package graph with all incoming packages & their deps
-    #[tracing::instrument(skip_all, fields(?lookup))]
-    fn update(&mut self, incoming: Vec<package::Id>, lookup: Lookup) -> Result<(), Error> {
+    #[tracing::instrument(skip_all, fields(lookup = %self.lookup))]
+    fn update(&mut self, incoming: Vec<package::Id>) -> Result<(), Error> {
```
jplatte: Looks like `add` and `update` are now identical, so we don't need `update` as a separate fn anymore, right?

```suggestion
    pub fn add(&mut self, incoming: Vec<package::Id>) -> Result<(), Error> {
```

moss/src/registry/transaction.rs
```diff
@@ -95,23 +81,23 @@ impl Transaction<'_> {
...
             }
             items = next;
         }
 
         Ok(())
     }
 
-    #[tracing::instrument(skip_all, fields(?check_id, check_name))]
-    fn update_step(&mut self, check_id: package::Id, next: &mut Vec<package::Id>, lookup: Lookup) -> Result<(), Error> {
+    #[tracing::instrument(skip_all, fields(%check_id, check_name))]
```
jplatte, line 94: Why change to `Display` formatting?

tarkah: It's cleaner, we get a non-debug value, so `some-id` vs `Id("some-id")` which is much nicer for structured logging purposes. If we eventually use something like `splunk`, it'd much easier to filter on `"some-id"` than the debug implementation

moss/src/registry/transaction.rs
tarkah: As per below, display implementation is much preferred for structured logging

tarkah:
Thanks @jplatte! I've addressed some of your review & left comments on a couple others

moss/src/registry/transaction.rs
jplatte: I removed the `Id()` from the debug format in the PR where I added this span ;)

moss/src/registry/transaction.rs
jplatte: That doesn't make sense to me here, Display and Debug are going to be the same as currently derived, no?

moss/src/registry/transaction.rs
tarkah: We don't need debug to output the same as display, that's why we have both implementations. We should be using `%...` to coerce display if that's how we want it logged, not changing the behavior of debug.

moss/src/registry/transaction.rs
tarkah: It's just misleading. Debug _shouldn't_ be the same as Display and in most cases isn't. Even if its just a string, it print's escape sequences and quotes. Very rarely do you want to log that value instead of the actual display format for k/v metadata and most of the time its because your _don't_ have a display implementation so you're forced to use the debug. There are definitely valid instances where you want to see the debug format of a string & its escape sequence when debug logging, but that's definitely not the case with these bits of k/v metadata.

jplatte: Well this isn't a string, it's an enum. I can see your point for the IDs, but I really dislike strum's display derive that just makes Display = Debug.

moss/src/registry/transaction.rs
jplatte: I don't think the `Id()` is useful in the Debug output, independently of what we do here. But I'll concede about using Display for IDs in tracing fields. It does make more sense in some ways. I just like the quotes around the string in the stdout text output, but I guess that is not what we should be optimizing for here (and with some luck we can have both soon enough).

ermo (approved):
Squash-merging this as is. Remaining nits can go in a follow-up PR.

PR #515, opened 2025-06-17
boulder/moss: add version flag

### LINKED ISSUE #233: moss + boulder: Streamline version info commands/options/flags
**Proposal**

Enable `version`, `-V` and `--version` as the supported arguments, so we increase the likelihood for people to get a match if they're looking for version info. Note that `rustc` supports `--version` and `-V`, so supporting these argument is part and parcel in the ecosystem.

Additionally, make `moss` and `boulder` print out version information as `# <executable> version (git hash)` if an invocation contains commands in addition to `-V/--version`, such as `boulder --version build` or `moss --version install (...)`.

This will enable infrastructure tooling to always call `moss` and `boulder` with the `--version` argument, which is a benefit when checking logs for which versions were used build/install packages, which in turn means that operationally, canary builds will also serve as a way to ensure that the deployed versions are the ones that we expected to be deployed.

**Context**

Currently, `moss` and `boulder` don't support the same arguments when trying to get version info:

    $ moss -V
    error: unexpected argument '-V' found
    
    Usage: moss [OPTIONS] [COMMAND]
    
    For more information, try '--help'.
    $ moss -v
    moss 0.1.0 (931a4f5)
    $ moss --version
    moss 0.1.0 (931a4f5)
    $ moss version
    moss 0.1.0 (931a4f5)

    $ boulder -V
    boulder 0.1.0 (931a4f5)
    $ boulder -v
    error: unrecognized subcommand 'version'
    
    Usage: boulder [OPTIONS] <COMMAND>
    
    For more information, try '--help'.
    $ boulder --version
    boulder 0.1.0 (931a4f5)
    $ boulder version
    error: unrecognized subcommand 'version'
    
    Usage: boulder [OPTIONS] <COMMAND>
    
    For more information, try '--help'.

HigherOrderLogic:
Resolves #233.

### DIFF
--- a/boulder/src/cli.rs
+++ b/boulder/src/cli.rs
@@ -37,6 +37,8 @@ pub struct Global {
         global = true
     )]
     pub verbose: bool,
+    #[arg(short = 'V', long, default_value = "false", global = true)]
+    pub version: bool,
     #[arg(long, global = true)]
     pub cache_dir: Option<PathBuf>,
     #[arg(long, global = true)]
@@ -63,6 +65,11 @@ pub enum Subcommand {
 pub fn process() -> Result<(), Error> {
     let args = replace_aliases(std::env::args());
     let Command { global, subcommand } = Command::parse_from(args.clone());
+
+    // Prints the cli's information about version at startup
+    if global.version {
+        println!("boulder {}", tools_buildinfo::get_full_version());
+    }
 
     if let Some(dir) = global.generate_manpages {
         fs::create_dir_all(&dir)?;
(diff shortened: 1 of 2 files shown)


jplatte (commented):
It seems pretty unusual to me that you would be able to combine `--version` with other flags. I am used to `--help` and `--version` being flags that supress all other flags and options. Admittedly, I had only done this with `--help` before, but I just tried `git --version checkout HEAD~` and `cargo --version update` and both only printed the version and exited.

To be clear: Apart from that, this looks good :)

ermo (approved):
This is implemented exactly as the issue outlines and per Reilly's suggestion.

Thanks.

PR #516, opened 2025-06-20
moss: Improve error output for `moss index`

ermo:
- Use fs_err as fs to improve error output where it was missing
- Add stone::reader::Error variant showing the path at the call-site
  This helps identify the culprit when indexing fails due to a checksum error.
  
  NB: This is meant to be a quick-and-dirty job for now -- in production, this needs to be "nicer".


jplatte (approved):

ermo:
Closing as superseded by #520 and #521

PR #523, opened 2025-06-22
Add --output-dir option to moss index

jplatte:
I have not tested the actual option, all the testing I've done can be seen in this PR (`mod tests`).

### DIFF
--- a/moss/src/cli/index.rs
+++ b/moss/src/cli/index.rs
@@ -8,7 +8,7 @@ use std::{
     time::Duration,
 };
 
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use clap::{ArgMatches, Command, arg, value_parser};
 use fs_err as fs;
 use moss::{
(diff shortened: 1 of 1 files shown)


moss/src/cli/index.rs
```diff
@@ -25,10 +25,18 @@ pub fn command() -> Command {
         .visible_alias("ix")
         .about("Index a collection of packages")
         .arg(arg!(<INDEX_DIR> "directory of index files").value_parser(value_parser!(PathBuf)))
+        .arg(
+            arg!(-o --"output-dir" [output_dir] "directory to write the index.stone to (defaults to INDEX_DIR)")
```
ermo: I like that you are calling it `index.stone`. This is technically more correct than the `stone.index` we currently use.

I might even be tempted to suggest that we begin naming the repo index `repo-index.stone` :sweat_smile:

ermo (commented):
I think we should standardise on names here:

- `pool_dir`: The directory containing .stones, potentially divided into sub-directories like vessel does
- `index_dir`: The directory containing the index of the the .stone files in the `pool_dir`

By default, `index_dir` will be equal to `pool_dir`. If `-o/--output` is present, that sets the index_dir wherein stone.index (or whatever name we come up with later on) is stored?

moss/src/cli/index.rs
jplatte: Oops :)

I actually don't think this is better though - it's better if `.stone` always means: "this is a package".

ermo (approved):
LGTM, thanks!

PR #536, opened 2025-07-01
github/cargo: Some compilation changes

ReillyBrogan:
Build dependencies with optimizations. Given that dependency compilation happens very infrequently this gives us most of the benefit of optimizations without increasing build times in the most common case of a developer rebuilding the project after making code changes (but not dependency changes).

Also bring back compressed debug symbols as an alternative for shrinking the binary with "line-tables-only". This shrinks the binary further than line-tables-only but without interfering with the ability to attach a debugger. To get back some of the lost linker time switch to using the mold linker.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,19 @@
+[target."x86_64-unknown-linux-gnu"]
+# - Compressing debug symbols with zstd shrinks the dev profile boulder binary from 206.03MB to 81.44MB, a 124.59MB
+# or ~60% savings. It doesn't affect the binary size for packaging builds since we strip those, but the debug symbols
+# are reduced in size from 113.16MB to 34.63MB.
+# - The new symbol mangling format (https://doc.rust-lang.org/rustc/symbol-mangling/v0.html) improves the backtrace
+# shown by RUST_BACKTRACE=1 and other debug utilities. It should also be helpful once we have ABI reports. Upstream
+# hasn't switched to it yet by default due to stable distros not having new enough tools, but that doesn't matter for us
+rustflags = [
+    "-Clink-arg=-fuse-ld=mold",
+    "-Clink-arg=-Wl,--compress-debug-sections=zstd",
+    "-Csymbol-mangling-version=v0",
+]
+
+[target."aarch64-unknown-linux-gnu"]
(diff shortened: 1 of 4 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,19 @@
+[target."x86_64-unknown-linux-gnu"]
+# - Compressing debug symbols with zstd shrinks the dev profile boulder binary from 206.03MB to 81.44MB, a 124.59MB
+# or ~60% savings. It doesn't affect the binary size for packaging builds since we strip those, but the debug symbols
+# are reduced in size from 113.16MB to 34.63MB.
+# - The new symbol mangling format (https://doc.rust-lang.org/rustc/symbol-mangling/v0.html) improves the backtrace
+# shown by RUST_BACKTRACE=1 and other debug utilities. It should also be helpful once we have ABI reports. Upstream
+# hasn't switched to it yet by default due to stable distros not having new enough tools, but that doesn't matter for us
+rustflags = [
+    "-Clink-arg=-fuse-ld=mold",
```
jplatte, line 9: I don't think we should be configuring mold on a per-project basis, as this requires anybody who wants to contribute to have it installed, increasing friction. This should be a user-wide config that we merely recommend IMO. By the way, [wild](https://github.com/davidlattimore/wild) is faster still in my experience :)

Cargo.toml
```diff
@@ -117,10 +117,9 @@ wildcard_imports = "warn"
 # has false positives in indicatif usage
 literal_string_with_formatting_args = "allow"
 
-[profile.dev]
-# makes linking faster, enables everything needed for backtraces
-# for attaching an interactive debugger, use another profile
-debug = "line-tables-only"
+# Build dependencies (but not workspace members) in release mode
+[profile.dev.package."*"]
+opt-level = 3
```
jplatte, line 122: I don't like this. Yes, full rebuilds don't happen _that_ often but for example when somebody uses nightly, they might upgrade their toolchain and clean out cargo target dirs every week or so (I personally use nightly and update every few weeks). I'd rather we enable optimizations for specific dependencies that we know to matter for runtime performance (if that is the problem being solved here - I haven't had issues with debug builds being too slow myself).

PR #539, opened 2025-07-19
cli/search_file: Ensure /usr/ prefix searches still succeed

joebonrichie:
the db doesn't contain the /usr/ prefix of the file, strip any combination of it to ensure queries such as /usr/bin/nano, /bin/nano and r/bin/nano still succeed.


moss/src/cli/search_file.rs
```diff
@@ -28,7 +28,18 @@ pub fn command() -> Command {
 }
 
 pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error> {
-    let keyword = args.get_one::<String>(ARG_KEYWORD).unwrap();
+    let mut keyword = String::from(args.get_one::<String>(ARG_KEYWORD).unwrap());
+
+    // moss db doesn't record the /usr/ prefix so strip any combination of it
```
ermo, line 31: If this was made into a separate search expression formatter function that takes a string and returns a string, then test cases could be made to demonstrate correctness for the function.

ermo (changes requested):
Please also run `just fix`on the code (as per the PR clippy lints).

ermo (approved):
Might as well get this in.

Issue #496, opened 2025-05-31
boulder: Updates to Python providers and depends


ReillyBrogan:
Per discussion in Matrix it has been decided that versioned python providers and depends (eg `python(setuptools(3.11))`) adds significant complexity that we are not yet ready for and little benefit to us at this time. While we would like to support co-installation of multiple major Python versions at once it is not something we can do at this time (due to the lack of package removal) and so the support will be removed until we are ready to properly support it.

Additionally review of PEP-503 (https://peps.python.org/pep-0503/#normalized-names) shows that our handling of such providers and dependencies names was incorrect. We need to apply all of the following rules to each python library name:
- Lowercase it
- Replace `.` with `-`
- Replace `_` with `-`

There is also no real point in inserting the non-normalized names into either dependencies or provides, we should just apply the normalization rules consistently and have that be the only name we care about.

PR #546, opened 2025-07-30
boulder: Drop versioned python deps

ReillyBrogan:
I just finished rebuilding all of the python packages that have versioned dependencies, now it's time to drop versioned providers.

Part of https://github.com/AerynOS/os-tools/issues/496


ermo (approved):
Let the unversioning commence.

PR #547, opened 2025-08-01
boulder: Disallow git tags and branches in upstreams to ensure reproducible builds

### LINKED ISSUE #109: boulder: Disallow using git tags in upstreams
The context here is that this is the easiest way to ensure that we get reproducible builds, given that projects sometimes silently repoint tags to different commit refs.

By disallowing this up front (via a `boulder lint` pass), we will never run into a situation where we have this happens to us in the first place.

Bonus points for ensuring the error message reads well, so packagers don't end up complaining/asking why boulder doesn't support tags.

barne856:
Closes #109 

Hopefully this is close to what the thinking was. I've placed the error formatting function at the module level for easy maintenance and reuse in tests. Let me know if you'd prefer a different location. And I'm certainly open to feedback on the exact messaging for the error.

### DIFF
--- a/crates/stone_recipe/src/lib.rs
+++ b/crates/stone_recipe/src/lib.rs
@@ -22,6 +22,16 @@ fn is_valid_sha1(s: &str) -> bool {
     s.len() == 40 && s.chars().all(|c| c.is_ascii_hexdigit())
 }
 
+fn format_git_ref_error(ref_id: &str) -> String {
+    format!(
+        "{} '{ref_id}'.",
+        "Using git tags or branches in upstreams is disallowed. \
+This helps ensure reproducibility since git tags and branches are mutable \
+references that can be updated to point to different commits over time. \
+Please use the full 40-character commit SHA instead of"
+    )
+}
+
 pub fn from_slice(bytes: &[u8]) -> Result<Recipe, Error> {
     serde_yaml::from_slice(bytes)
 }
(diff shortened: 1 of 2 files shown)


barne856:
> With this change, the documentation [here](https://aerynos.dev/packaging/recipes/upstreams/) will need to be updated to reflect the new requirement.

Since the message is a bit wordy, we could make it more concise and move the detailed explanation to the docs. Then include a link to it from the error message?

barne856:
Just opened https://github.com/AerynOS/dotdev/pull/29 I'm thinking we should rename `ref` to `hash` for git upstreams since we are essentially disallowing all `refs` with this change.

crates/stone_recipe/src/lib.rs
```diff
@@ -18,6 +18,20 @@ pub mod macros;
...
+    s.len() == 40 && s.chars().all(|c| c.is_ascii_hexdigit())
+}
+
+fn format_git_ref_error(ref_id: &str) -> String {
+    format!(
+        "{} '{ref_id}'.",
+        "Using git tags or branches in upstreams is disallowed. \
+This helps ensure reproducibility since git tags and branches are mutable \
+references that can be updated to point to different commits over time. \
+Please use the full 40-character commit SHA instead of"
```
ermo: Perhaps it would be a nice UX improvement to actually attempt to fetch and present the correct SHA sum here?

Said fetch/translate function could then also be used in `boulder recipe new/update`.

ermo (dismissed):

barne856:
> Sounds reasonable. The consequence is that we will be using hash for both git URIs and plain URIs, which is not necessarily a bad thing.

An alternative is to be more specific about they type of hash? Plain upstreams seem to use sha256 while git upstreams use sha1?

```
upstreams:
    - git|uri:
        sha1: $sha1
```

```
upstreams:
    - uri:
        sha256: $sha256
```

ermo:
> > Sounds reasonable. The consequence is that we will be using hash for both git URIs and plain URIs, which is not necessarily a bad thing.
> 
> An alternative is to be more specific about the type of hash? Plain upstreams seem to use sha256 while git upstreams use sha1?
> 
> ```
> upstreams:
>     - git|uri:
>         sha1: $sha1
> ```
> 
> ```
> upstreams:
>     - uri:
>         sha256: $sha256
> ```

One of the goals of the recipe format is to be as low-ceremony as possible. If we can get away with using `hash` and have the tooling auto-detect the format, that cuts down on the mental overhead of having to care which hash algorithm we're dealing with...?

There's the additional consideration that git is slowly moving to sha256sum hashes as well. In light of that, it might be "enough" to match on the length...?

crates/stone_recipe/src/lib.rs
ermo: fixed in newest version, so resolving.

Issue #312, opened 2024-08-09
moss: Work on `moss state info` UX


ermo:
Given the `moss state` mechanism appears to be working as expected, it might be wise to begin looking at how to make the feature more discoverable and more user friendly.

Iterms to consider:
- Add the ability to label an fstx with a `-C/--comment` flag maybe?
- Add the ability to be shown a list of transactions with a `-L/--list-transactions` flag maybe?
- Add the ability to `diff` two states, meaning that moss would show what would change in terms of the installed set of packages prior to activating a different fstx state. Could work with a `--show-diff` flag maybe?

Issue #114, opened 2023-12-23
Multi-version Support?


trans:
Does moss support multiple version installs (either Gobo style or Nix style)?

ermo:
Slots are planned so it becomes possible to have e.g. clang-16 and clang-17 installed side-by-side if need be.

However, in general, the model we're going for is to ensure that a "versioned tree" (think a git ref of a recipe tree) is internally consistent in terms of build-time ABI as in "transitive closure of the ABI graph".

This in turn implies that, for any given "repo-release" our builders will construct build queues until the ABI transitive closure property can be proven to hold for all the available binary artefacts built from the given recipe repo, which in turn implies that that the ABI-related property of "package versions" that you might be used to, will not be interesting to us from a packaging perspective.

If you want/need the sort of system where package versions matter, both Gobo and Nix are better options than Serpent.

As an aside, I contributed a bit to Gobo few years back, so I know what you mean by the version thing in that regard.

PR #555, opened 2025-08-08
moss/cli: Add state query and state diff subcommands

joebonrichie:
Query allows to view the selection list for a particular state

Diff allows to view the difference between state A and state B

Part of #312

### DIFF
--- a/moss/src/cli/state.rs
+++ b/moss/src/cli/state.rs
@@ -2,12 +2,15 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+use std::{cmp::Ordering, collections::HashMap};
+
 use chrono::Local;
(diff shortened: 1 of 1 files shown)


ermo (changes requested):
Needs clippy lints fixed.

In addition, I think it would be wise to be clear which column belongs to which transaction, particularly if it is a long list of changes that scrolls original cli invocation text off the screen.

So maybe a legend at the bottom, showing text like "diff from fstx #114 -> fstx #112" or "diff from fstx #112 -> fstx #114" in response to the cli input?

Would probably also be wise to show a screenshot in the cover letter in the PR (not the commit).

ermo (approved):
Let's see where this takes us. Thanks @joebonrichie

PR #560, opened 2025-08-09
do not try to resolve git upstream hashes if they are already provided

barne856:
fix for this bug: https://gist.github.com/ermo/4f4f5eddb93780b9c14d072abd494eec


ermo (approved):
Tested to work on the previously failing build.

LGTM, thanks!

Issue #556, opened 2025-08-08
boulder enhancement for packaging


bhh32:
# Boulder Packaging Tool Enhancement Suggestions

## Reasoning
I was re-packaging all of the COSMIC packages getting them up-to-date. I realized that if the multi-step approach was built into the single `boulder recipe update...` command it would go much faster.  
  
Also, while building one of the packages I had a failure due to the updated version needing a new build dependency. Instead of having a good error message that showed why the build failed, I had to sift through all of the compiling messages from the output to find what had went wrong and what package I needed to add to the stone.yaml file to make it successfully build.  
  
## Solution
TODO: I will look through the boulder code and put specific updates that I would make to improve this particular use-case of the tooling and to solve the error issue when a build fails mid compilation.  
  
Once I have figured out what I would change/add, I will post it within this issue to discuss whether or not I should try implementing the solution and create a PR for the tooling.

bhh32:
I don't mind flipping it to build instead of no_build. I can make something in the next day or two that others can test to see if they like it.

ermo:
To me, this issue is about "bringing home to boulder" the additional "layer" of composable features currently implemented via shell and  `just` recipes.

The first feature is about auto-building a recipe after updating it.

The second feature is about choosing which repository a newly built recipe should be moved to.

Hence it follows that the way to organise this, is to first ensure that the `boulder build --mv-to-repo=<some valid repo>` (which is a generic implementation of the previous --local argument) actually works, and _then_ making it so that `boulder recipe new --build --mv-to-repo=<somerepo>` works as well.

It is common practice for people to do stack updates in separate repos that are _not_ their primary local repo. Hence, this way of approaching the feature is indicative of how long-time packagers actually use boulder.

For now, everything else but these two features is out of scope. If we want to expand on this later on, we make a new issue and a new PR.

bhh32:
@ermo I understand the ask now. It's definitely within reason to make this change. I can work on moving things around and see what people think. Is there anyone else currently working on the `boulder build` command? I don't want to cause any merge conflicts or accidentally overwrite any changes. Basically, I'd like to coordinate efforts to ensure we're jiving on our changes.

ermo:
Nobody is working on boulder build.

One stretch goal feature you could look at (no expectations) is a `boulder build --mv-to-repo=<somerepo> --re-lndex` flag.

The default would be to not re-index after a build and move (to keep with current defaults).

[...]

PR #561, opened 2025-08-10
Boulder recipe update additions

bhh32:
## What Was Updated
Updated the `boulder recipe update` command with `build` and `local` flags. These new flags allow for a package maintainer to automatically update the stone.yaml file, build the package, and move it to the local repository. Discussion is documented in #556.

## Behavior
The default behavior is the same as before the changes. This means without these two new flags packagers will have the same experience they did before the changes.

If the packager only uses the `--build` flag the package stone.yaml file is updated and the package is automatically built on a successful update. The package is not moved to the local repository.

However, if both `--build` and `--local` are given, the stone.yaml file is updated, the package is built, and the built package is automatically moved to the user's local repository on a successful build process.

The end result looks like the following:  
  
Screenshot_2025-08-09_23-52-10

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @ikeycode @ermo @tarkah
+* @ikeycode @ermo @AerynOS/tooling:core
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,6 +404,7 @@ dependencies = [
  "stone",
  "stone_recipe",
  "strum",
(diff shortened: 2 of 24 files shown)


bhh32:
@NomadicCore, almost. I was trying to clean up so merge conflicts when I rebased last night. Hopefully will have it fixed and ready to go today or tomorrow.

bhh32:
Closing PR to open a new one with cleaner commit history.

PR #562, opened 2025-08-10
Boulder: Refactor boulder git hash resolution

barne856:
Per discussion on matrix, this refactors git upstream hash resolution to use `git rev-parse` on cloned repositories instead of `git ls-remote` during build initialization.

This is a backwards compatible change but should allow for more flexibility in the `ref` specified in the `stone.yaml`. In particular any ref or hash that `git rev-parse` can parse should resolve to the full 40 character hash.

I've include a suite of tests for both the new `resolve_git_ref` and `update_git_upstream_refs` that test various refs and hashes in different formats. I think it is comprehensive enough, but let me know if you see any obvious holes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,6 +404,7 @@ dependencies = [
  "stone",
  "stone_recipe",
  "strum",
+ "tempfile",
  "thiserror 2.0.12",
  "thread-priority",
  "tokio",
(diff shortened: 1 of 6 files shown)


jplatte (approved):
Nice, thanks for writing tests!

I wonder whether we could also do `rev-parse HEAD` since we the relevant ref is always the checked-out one? But this way is probably more robust.

PR #570, opened 2025-09-03
boulder: Add '--build', '--mv-to-repo=', and '--re-index' flags

bhh32:
# Purpose
This PR is the cleaned up, ready to merge, work for #561.

## Flag Usage
The `--build` flag is only to be used with the variations of the `recipe` sub-command. It's purpose it to automatically run the build sub-command after the recipe sub-command's process has successfully completed.
  
Example:  
```bash
boulder recipe update stone.yaml -w --build
```  
  
The `--mv-to-repo=` flag is restricted to only be used when either the `build` sub-command or `--build` flag has been used. It didn't make sense to allow it to be used with the other sub-commands or if something wasn't being newly built. It's purpose is to automatically all newly built .stone files to the designated - moss registered - repo. Contrary to the `just mv-local` command, it does not automatically re-index the given repository.
  
Examples:
```bash
boulder build --mv-to-repo=local
boulder recipe bump --build --mv-to-repo=local
```  
  
Finally, the `--re-index` flag is given to automatically re-index the repo that was given to the `--mv-to-repo=` flag. Therefore, the `--re-index` flag requires the `--mv-to-repo=` flag.  
  
Examples:  
```bash
boulder build --mv-to-repo=local --re-index
boulder recipe update --build --ver 1.0.0 --upstream "git|deadbeef" stone.yaml -w --re-index --mv-to-repo=local
```


boulder/src/cli.rs
```diff
@@ -120,10 +165,166 @@ pub fn process() -> Result<(), Error> {
...
+        Some(Subcommand::Build(command)) => {
+            match build::handle(command, env) {
+                Ok(_) => {
+                    if let Some(repo) = global.mv_to_repo {
+                        // Check to see if the repo is in moss
+                        let moss_cmd = std::process::Command::new("moss")
+                            .args(["repo", "list"])
+                            .stdout(std::process::Stdio::piped())
+                            .output()
+                            .expect("Couldn't get a list of moss repos");
```
tarkah, line 177: We can use `moss` APIs directly from Rust, we shouldn't need to invoke it as a separate executable.

boulder/src/cli/recipe.rs
```diff
@@ -116,7 +124,34 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
...
+        && let Ok(_) = run_cmd
+    {
+        use std::process::{Command as Cmd, Stdio};
+
+        let mut boulder_build = Cmd::new("boulder")
+            .args(["build", "-u", "stone.yaml"])
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
+            .spawn()
+            .expect("Failed to run boulder build command!");
```
tarkah, line 139: Same here, we can just call a common function directly vs re-invoking `boulder` from `boulder`.. `cli::build::handle(..)`

boulder/src/cli.rs
```diff
@@ -120,10 +165,166 @@ pub fn process() -> Result<(), Error> {
...
+                let moss_cmd = std::process::Command::new("moss")
+                    .args(["repo", "list"])
+                    .output()
+                    .expect("Couldn't get a list of moss repos");
+
+                let repos = String::from_utf8(moss_cmd.stdout).expect("Could not get the repo list from moss");
+
+                let mv_repo = repos
+                    .lines()
+                    .filter_map(|line| {
```
tarkah, line 266: Can we factor out the `mv-to-repo / re-index` logic and call the single function from both `build` and `recipe` branches? This can all just be handled within `cli::build` since its a side-effect of the build and when calling `cli::build::handle` from `recipe`, we can just forward these arguments.

boulder/src/cli/recipe.rs
```diff
@@ -29,6 +29,14 @@ use url::Url;
...
 #[command(about = "Utilities to create and manipulate stone recipe files")]
 pub struct Command {
+    #[arg(
+        long,
+        required = false,
+        default_value_t = false,
+        global = true,
+        help = "Build the recipe after successful completion of the subcommand"
+    )]
+    pub(crate) build: bool,
```
tarkah, line 39: `cli` is part of `main` which isn't a library so `pub(crate) and pub` are semantically the same. Regardless I think having all fields as `pub` for these CLI parsing structs is fine.

boulder/src/cli.rs
bhh32: My thought here is I'd like to refactor both boulder and moss to have a shared library crate. This way they both can call shared logic and not rely directly upon each other. To me, and I could be wrong, invoking moss directly here reduces the refactoring later if that were to become a thing. It also marks where the refactoring needs to happen.

boulder/src/cli/recipe.rs
bhh32: Yes, I can definitely refactor this.

boulder/src/cli.rs
bhh32: Yep, absolutely. I can make this change for sure.

boulder/src/cli/recipe.rs
bhh32: Sounds good! I'll replace it with `pub`.

boulder/src/cli.rs
tarkah: Refactoring is much easier w/ the help of `rustc` / `rust-analyzer` :) If we broke these interfaces, we'd get no compilation errors here. We can also build a much better API if we handle this natively in `moss` lib, such as `moss::repository::Manager::get(repo: &str) -> Option<Repository>`.

PR #572, opened 2025-09-03
boulder/data: Add macros for perl Module::Build

joebonrichie:
Module::Build is an alternative to ExtUtils::MakeMaker from the stdlib that some projects use.

### DIFF
--- a/boulder/data/macros/actions/perl.yaml
+++ b/boulder/data/macros/actions/perl.yaml
@@ -1,12 +1,31 @@
 actions:
-
-    # NOTE: If required we can add support for the deprecated Module::Build buildsystem in the future.
-    #       However, let's just support ExtUtils::MakeMaker for now to vastly simply these macros.
-
     - perl_setup:
-        description: Setup perl with ExtUtils::MakeMaker from stdlib
+        description: Create a Makefile with ExtUtils::MakeMaker from stdlib to be used with %make and %make_install macros
         command: |
             perl Makefile.PL PREFIX="%(prefix)" NO_PACKLIST=1 NO_PERLLOCAL=1 INSTALLDIRS=vendor DESTDIR="%(installroot)"
         dependencies:
             - binary(perl)
 
+    - perl_module_setup:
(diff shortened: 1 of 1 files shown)


ermo:
@ReillyBrogan please give this a look?

boulder/src/draft/build/perl.rs
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: Copyright © 2025 Aeryn OS Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
+pub mod perl_extutils_makefile {
+    use crate::draft::File;
```
tarkah: Nit: Redundant naming `perl`, we're already in the `perl `module.. `perl::extutils_makefile::phases`

boulder/src/draft/metadata/metacpan.rs
```diff
@@ -0,0 +1,46 @@
...
+pub fn source(upstream: &Url) -> Option<Source> {
+    let re = Regex::new(
+        r#"^https://cpan\.metacpan\.org/authors/id/[A-Z]/[A-Z]{2}/[A-Z0-9]+/([A-Za-z0-9._+-]+-\d+(?:\.\d+)*)(?:\.tar\.(?:gz|bz2|xz)|\.zip)$"#
+    ).ok()?;
+
+    let captures = re.captures(upstream.as_str())?;
+
+    let module = captures.get(1)?.as_str().to_owned();
+    let parts: Vec<&str> = module.split('-').collect();
+
```
ermo, line 19: nit: It might be nice if you could add a common perl module `name-version` string here for reference as a comment?

Is this something that we could add tests for?

ermo (commented):
A couple of questions.

boulder/src/draft/metadata/metacpan.rs
tarkah: Unit test would definitely be appreciated :+1: Non blocking though

ermo: Captured the unit test ask as an issue here: https://github.com/AerynOS/os-tools/issues/591

PR #574, opened 2025-09-09
moss: Parallel state verify with rayon iterators

### LINKED ISSUE #390: moss: Explore making `moss state verify` able to use multiple threads when verifying

jplatte:
Alternative to #421.
Resolves #390.

This avoids leaking anything about the execution order into the output. It could however be quite hard to extend for anybody not strongly familiar with advanced iterator combinators (and rayon adds a little bit of complexity of its own).

I expect this to be no slower than #421, potentially faster. I have not verified this code in any way beyond making sure it compiles. I would love to find a less complex-looking solution to this, and maybe there is (now that I'm thinking about it, I vaguely remember seeing a crate that helps with fallible parallel list building using rayon).

### DIFF
--- a/moss/src/client/verify.rs
+++ b/moss/src/client/verify.rs
@@ -40,7 +40,6 @@ pub fn verify(client: &Client, yes: bool, verbose: bool) -> Result<(), client::E
             .push((package, file));
     }
 
-    let mut issues = vec![];
     let mut hasher = digest::Hasher::new();
 
     let pb = ProgressBar::new(unique_assets.len() as u64)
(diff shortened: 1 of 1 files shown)


jplatte:
So overall, what's your opinion on #421 vs. this?

joebonrichie:
LGTM, having these massive iterators is a bit blurgh, but, it's just using what was already there. Also, not having fragile lock usage is a plus.

PR #578, opened 2025-09-13
moss/cli: Remove state diff support for now

joebonrichie:
Rewrtten this a couple of times and continue to find edge cases not handled correctly. I would rather remove this feature for now till it's handled properly.


ermo (approved):
Agreed that it is probably better to not have this than have it be wrong.

PR #589, opened 2025-09-30
boulder: Add undefined-version ldflag

ReillyBrogan:
Unfortunately it is error prone to watch configure/cmake/meson logs to see which packages need `-Wl,--undefined-version` since if the flag doesn't work they'll just end up skipping using it. Since we do want version scripts to be applied if present let's just add the appropriate flag to allow for version scripts with undefined or duplicate symbols.

### DIFF
--- a/boulder/data/macros/arch/base.yaml
+++ b/boulder/data/macros/arch/base.yaml
@@ -121,6 +121,7 @@ defaultTuningGroups :
     - optimize
     - relr
     - symbolic
+    - version-allow-undefined
 
 tuning              :
     # A set of groups we can toggle from the "tune" key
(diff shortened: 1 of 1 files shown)


ermo (approved):
Not 100% happy with the naming, but not going to hold it back for that reason alone, since the functionality is genuinely useful.

ReillyBrogan:
> Not 100% happy with the naming, but not going to hold it back for that reason alone, since the functionality is genuinely useful.

If you or @joebonrichie can think of better naming by all means feel free to change it. All I care about is that `-Wl,--undefined-version` is added to the default LDFLAGS which this PR achieves. I don't actually foresee actually disabling it for packages.

PR #563, opened 2025-08-15
Moss install tracing

barne856:
## Summary

An initial attempt at adding tracing to `moss install` per #558. This does not close the issue and is meant to be a focused first PR to be reviewed and disscussed so we can eventually provide better observability and tooling integration for all of moss. I'm not very familiar with Rust's tracing crate so please do correct me if I'm doing anything unconventional, but I tried to base this off of the lichen-installer repo for reference.

The focus here is on providing progress events through tracing. The following items represent necessary (though not sufficient) work that remains:

- ensure we only use special formatting when isatty() is true for stdout
- instrument more functions, add more logging?
- Ikey mentioned this would producing a prototype for packagekit integration, but I think that work has already started? Not sure what if any effect this has on that work.

## What changed

Replaced the old `--debug` flag for `moss install` with a more flexible `--log` option that takes `<level>[:<format>][:<destination>]`. By default, format is `text` and destination is `stderr`. So you can do things like:

- `moss install --log debug foo` (debug to stderr)
- `moss install --log info:json:moss.log bar` (JSON to file)
- `moss install --log warn:text:stderr baz` (warnings in text format to stderr)

Created a new `tracing_common` crate to share logging utilities across the project. It has a logging module to handle tracing subscriber setup and a progress module with helpers for reporting the installation progress.

Added `#[instrument]` to the `handle`, `install`, and `resolve_input` functions in the `moss install` pipeline.

Emit progress events for package caching (overall progress for number of packaged cached, and per package progress for downloads and unpacking) and package installation (blitting and triggers per item).

Refactored some messaging to make sure we keep a clean separation between stdout and stderr:

- stdout: normal user messages and prompts
- stderr: only tracing output (when it's the destination)

When tracing ouput is a log file, ANSI colors are disabled so we don't have random color codes in the log file.

Also fixed a small bug? In `moss/src/package/render.rs` where it was writing source release twice instead of source release and build release for packages.

## Example Tracing Output

```bash
moss install --log info:json:log.txt zed
```

```json
{"timestamp":"2025-08-15T05:10:24.063689Z","level":"INFO","fields":{"message":"Package resolution completed","total_resolved":28,"missing_packages":28,"already_installed":0,"resolve_time_ms":"222"},"target":"moss::client::install","span":{"ephemeral":false,"pkgs":"[\"zed\"]","yes":false,"name":"install"},"spans":[{"name":"handle"},{"ephemeral":false,"pkgs":"[\"zed\"]","yes":false,"name":"install"}]}
[...]

Issue #558, opened 2025-08-09
Add sane, streamlined reporting capability via the `tracing` crate, w/ stdout = normal output and stderr = error output


ermo:
## Key Priorities

- Integrate os-info into blsforme / moss to handle rename of serpent-os identity to aerynos for ESP/os-release migration
    - Need: systemd socket activation in lichen-installer backend. Drop privileged-ipc / tools-api crap everywhere.
    - Also needed: polkit integration in backend for unix sockets by way of tonic connection handler or tower middleware.

### Jump in points for others

- moss needs to integrate tracing to handle all logging with correct levels, so we can get insights on whats being done internally, like outputs of triggers.
  (ensures better integration story down the road)
- Need to decorate/instrument functions, as well as specially add events for progress signals, work ongoing etc. We'll then add after a special JSON tracing output for scraping purposes to connect moss into other tooling.
- Reference for all works: https://github.com/AerynOS/lichen-installer

#### Context

> **Ikey Doherty**
> Add the relevant info/debug/event/instrument stuff 
> Then we can add a special json output thingy 
> This will allow us to actually emit errors and debug info in moss properly and have a log file for sessions 
> Give insights into failed triggers 
> (Like the failed triggers when run in chroot...)
> And also empower lichen to connect to progress events 
> Which in turn is producing a prototype for packagekit integration 
> (Therefore gnome-software/ cosmic-store)
> If someone's working on that *now* then when I return to lichen later I can hook it up there and voila progress signals for moss invocation 
> Which is awesome because old lichen just directly calls moss and suspends progress bars
> Which is utterly useless for when we do add gui frontend
> (...)
> **Ikey Doherty**
> So had some questions about new focus on porting to tracing crates
> Tldr we should only ever error via stderr otherwise tooling integration is Hard
> Whereas normal human ui elements go to stdout
> Likewise we should have meaningful exit codes
> And ensure we only use special formatting when isatty() is true for stdout
> Simple use case:
> `pkg=$(moss info somename 2> /dev/null)`
> This way only useful output is captured and exit code is relied on, or empty output.
> Likewise, eprintln acquires a lock for stderr which may interfere with progress/ message rendering in moss, so we totally avoided printing anything from child processes or errors 
> Including triggers.
> In reality, we should be using tracing heavily and reporting all of these things, then integrating higher up to interleave emission of tracing/ log messages of a minimum level to our existing output ,which could suspend progress bars and correctly render without breaking 
[...]

PR #590, opened 2025-09-30
moss: Add tracing to sync command

barne856:
Add tracing to the `moss sync` command per #558 following the same pattern established in the `install` command #563.

### DIFF
--- a/moss/src/cli/sync.rs
+++ b/moss/src/cli/sync.rs
@@ -17,7 +17,7 @@ use moss::{
 };
 use thiserror::Error;
 
-use tracing::{debug, info, info_span, instrument};
+use tracing::{Instrument, debug, info, info_span, instrument};
 use tui::dialoguer::Confirm;
 use tui::dialoguer::theme::ColorfulTheme;
 use tui::pretty::autoprint_columns;
@@ -152,7 +152,7 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         event_type = "progress_start"
     );
 
-    runtime::block_on(client.cache_packages(&synced))?;
+    runtime::block_on(client.cache_packages(&synced).in_current_span())?;
 
     timing.fetch = instant.elapsed();
     info!(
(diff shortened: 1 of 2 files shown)


moss/src/cli/sync.rs
```diff
@@ -123,8 +141,30 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
+    let cache_packages_span = info_span!("progress", phase = "cache_packages", event_type = "progress");
+    let _cache_packages_guard = cache_packages_span.enter();
+    info!(
+        phase = "cache_packages",
+        total_items = synced.len(),
+        progress = 0.0,
+        event_type = "progress_start"
+    );
+
     runtime::block_on(client.cache_packages(&synced))?;
```
jplatte: If we want the async stuff here, which runs on tokio's threadpool, to have the span above, the `.enter` is not enough. You'd have to call ~`.instrument(the_span)`~ `.in_current_span()` (from `tracing::Instrument`) on `cache_packages(...)`.

tarkah: ^^ 

```suggestion
    runtime::block_on(
        client.cache_packages(&synced)
            .instrument(span)
    )?;
```

tarkah (approved):
LGTM, though definitely address the feedback left by @jplatte before merging.

I think eventually we can define tracing in a much more structured way, where `event_type` and `phase` and other future things are strongly typed and maybe we provide our own tracing API to force us to define those things vs relying on `tracing` directly.

Also, eventually eventually I think we need to abstract away all tracing / human readable / progress output into some sort of trait and eliminate all direct usage of `println / tui::progress / tracing crate / etc` so we can lock down exactly how we are outputting stuff & make it easily configurable what type of output you want to see and in what format. 

Getting in some basic tracing initially is a great first step, thanks!

moss/src/cli/sync.rs
jplatte:
```suggestion
    runtime::block_on(
        client.cache_packages(&synced)
            .in_current_span()
    )?;
```

PR #595, opened 2025-10-17
boulder: Eliminate all lingering threads

tarkah:
After spending some time in `lldb` to figure out
what threads still exist when we enter the container, I found the following:

- zbus "blocking" wraps its async client and uses the `async-io` crate to `block_on` that client. This created a single global runtime thread that we had no control over. I've changed this to opt-in to its `tokio` feature so it won't spawn its own runtime, and modified `runtime::block_on` to allow things
to get called when no global (`runtime::init`) runtime is initialized by running it on a temporary runtime. In either case, all threads are now dropped.

- rayon spawns its own global runtime we have no control over. I've updated the `par_iter` invocation in the boulder call path to use a temporary runtime that we can drop

All threads are now dropped before we CLONE into the user namespace ("container") which should hopefully eliminate the risk of us sometimes deadlocking

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,18 +137,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-channel"
-version = "2.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "924ed96dd52d1b75e9c1a3e6275715fd320f5f9439fb5a4a11fa51f4221158d2"
-dependencies = [
- "concurrent-queue",
- "event-listener-strategy",
- "futures-core",
- "pin-project-lite",
-]
-
-[[package]]
 name = "async-compression"
 version = "0.4.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Cargo.lock
jplatte, line 1: Very nice side effect of reducing dependencies! :)

moss/src/runtime.rs
```diff
@@ -53,14 +53,26 @@ impl Runtime {
     }
 }
 
-/// Run the provided future on the current runtime.
+/// Run the provided future on the global runtime if its
```
jplatte:
```suggestion
/// Run the provided future on the global runtime if it's
```

moss/src/runtime.rs
```diff
@@ -53,14 +53,26 @@ impl Runtime {
...
+/// been initialized, otherwise it creates a temporary runtime
+/// to run the future on and then fully drops it
```
jplatte:
```suggestion
/// to run the future on and then fully drops it.
```

PR #596, opened 2025-10-21
moss: Add system model MVP

tarkah:
This adds an initial MVP of the long mooted system-model feature, which offers a very useful way to take advantage of the Versioned Repositories [Channel API PR](https://github.com/AerynOS/infra/pull/142).

## Imperative moss usage

We define "imperative" usage of moss as the existing workflow. Users issue consecutive commands to mutate the state of their system. During `sync`, the active state is the source of truth. If you were to `remove` a package, that new state is the source of truth.

## Declarative moss usage (`system-model`)

A system-model is a _declarative_ way to define the desired state of a system. A system model file explicitly defines the repositories & packages a user wants to `sync` their system against. During `sync`, the system-model file is the source of truth instead of the current active state.

`install`, `remove`, `state active`, etc can still be used to imperatively mutate the active state, but unless the desired changes are written to the system-model file, these changes are temporary and will be lost on the next `sync`. The inspiration for this behaviour is the way Nix works, but there are otherwise no parallels.

Users _opt in_ to using system-model by writing a `{root}/etc/moss/system-model.kdl` file.

## system-model.kdl example

A very simple implementation of defining repositories and packages via KDL is implemented and looks like this:

```kdl
repositories {
    volatile-stream {
        uri "https://build.aerynos.dev/stream/volatile/x86_64/stone.index"
        priority 1
    }
}
packages {
    bash
    nano {
      why "because vim scares me"
    }
    "pkgconfig(zlib)"
    "binary(cargo)"
}
```

KDL playground here: https://kdl.dev/play/

## TODO

There are things still left to do, like flesh out the full `system-model.kdl` format and ensure `moss` UX is consistent and accurate across all commands when `system-model` is active.

- [x] disable repo add command
- [x] disable repo remove command
- [x] `moss state export`
- [x] add "system-model in use" pre-amble

### DIFF
--- a/moss/src/cli/mod.rs
+++ b/moss/src/cli/mod.rs
@@ -13,6 +13,7 @@ use clap_mangen::Man;
 use moss::{Installation, installation};
 use thiserror::Error;
 use tracing_common::{self, logging::LogConfig, logging::init_log_with_config};
+use tui::Styled;
 
 mod boot;
 mod extract;
@@ -192,6 +193,10 @@ pub fn process() -> Result<(), Error> {
 
     let installation = Installation::open(root, cache.cloned())?;
 
+    if installation.system_model.is_some() {
+        print_system_model_warning(&installation);
+    }
+
     match matches.subcommand() {
         Some(("boot", args)) => boot::handle(args, installation).map_err(Error::Boot),
         Some(("extract", args)) => extract::handle(args).map_err(Error::Extract),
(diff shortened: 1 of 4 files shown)


ermo (commented):
Per our chat on how we communicate that this feature is not production ready, I think we would do well to add a moss preamble showing a warning or an info message to the user that system-model support is experimental.

ermo (approved):
LGTM. We can fix any style issues after the fact, as we need this in for testing and productisation.

PR #598, opened 2025-10-24
moss: Add tracing to remove command

barne856:
Add tracing to the `moss remove` command per #558.

Also changed progress message from "Installing" to "Blitting" in `blit_element()` for consistency across install/remove/sync operations.

### DIFF
--- a/moss/src/cli/remove.rs
+++ b/moss/src/cli/remove.rs
@@ -16,7 +16,7 @@ use moss::{
     registry::transaction,
     state::Selection,
 };
-use tracing::{debug, info, info_span, instrument, warn};
+use tracing::{debug, info, instrument, warn};
 use tui::{
     Styled,
     dialoguer::{Confirm, theme::ColorfulTheme},
@@ -88,7 +88,6 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         "Package resolution for removal completed"
     );
 
-    debug!(count = removed.len(), "Full package list for removal");
     for package in &removed {
         debug!(
             name = %package.meta.name,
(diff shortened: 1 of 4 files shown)


ermo:
@tarkah Please take a look at this?

moss/src/cli/remove.rs
```diff
@@ -74,6 +80,25 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
 
+    timing.resolve = instant.elapsed();
+    info!(
+        total_packages = removed.len(),
+        packages_to_remove = removed.len(),
+        resolve_time_ms = timing.resolve.as_millis(),
+        "Package resolution for removal completed"
+    );
+
+    debug!(count = removed.len(), "Full package list for removal");
```
tarkah: I think we can remove this `debug`, its already captured by the above ` info`. We can just debug log the individual packages as you do after this.

moss/src/cli/remove.rs
```diff
@@ -91,8 +116,19 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         return Err(Error::Cancelled);
     }
 
+    instant = Instant::now();
+
+    let removal_span = info_span!("progress", phase = "removal", event_type = "progress");
+    let _removal_guard = removal_span.enter();
+    info!(
+        phase = "removal",
```
tarkah: `phase` doesn't need to be redefined, you've added it at the span level already.

tarkah: I'm not sure these "progress" traces give us much at this layer? We already track the total vs blit time on the final log. Any progress related to blitting or the implementation of creating a new state should be handled within the client, not at this layer.

barne856: Thanks, I agree moving these into the client make more sense. I've refactored the `install` and `sync` commands as well to be consistent and removed the unnecessary phase definitions in those spans as well.

moss/src/client/mod.rs
```diff
@@ -309,7 +323,17 @@ impl Client {
...
-        }
+        };
+
+        info!(
+            duration_ms = timer.elapsed().as_millis(),
+            items_processed = selections.len(),
+            progress = 1.0,
+            event_type = "progress_completed",
+        );
+        drop(_state_guard);
```
tarkah, line 334: Nit: No need to drop the guard manually, you're at the end of the scope and it'll be dropped

PR #601, opened 2025-10-26
boulder: Mount ccache config if exists

ReillyBrogan:
If `/etc/ccache` exists on the host system, mount it into the container. This allows for overriding the ccache config.

### DIFF
--- a/crates/container/src/lib.rs
+++ b/crates/container/src/lib.rs
@@ -81,6 +81,21 @@ impl Container {
         self
     }
 
+    /// Create a read-only bind mount only if the `host` path exists
+    pub fn bind_ro_if_exists(mut self, host: impl Into<PathBuf>, guest: impl Into<PathBuf>) -> Self {
+        let source = host.into();
+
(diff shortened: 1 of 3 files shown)


crates/container/src/lib.rs
```diff
@@ -56,22 +56,26 @@ impl Container {
     }
 
     /// Create a read-write bind mount
```
ermo, line 64: Please document what the optional bool is supposed to be used for and when?

crates/container/src/lib.rs
```diff
@@ -56,22 +56,26 @@ impl Container {
...
+            self.binds.push(Bind {
+                source: host.into(),
+                target: guest.into(),
+                read_only: false,
+            });
+        }
         self
     }
 
     /// Create a read-only bind mount
```
ermo, line 74: Same as above, please document what the optional bool is supposed to be used for and when?

ermo (dismissed):
Seems useful. Just a couple of nits.

PR #603, opened 2025-10-27
boulder: Ensure recipe release is positive

ermo:
Output in case of `release: 0`:

```
Error: build: build recipe: recipe: value: release must be > 0 (found 'release: 0')
```


tarkah:
We can add this check to where we load the recipe file:

https://github.com/AerynOS/os-tools/blob/main/boulder%2Fsrc%2Frecipe.rs#L33

PR #605, opened 2025-10-31
Task/tx cache speedup

tarkah:
## Selections cache

When a package is added to a TX, we need to ensure future
packages that depend on its providers resolve to that selected
package and not something else in the registry. Before we
queried the registry always and then checked to see if it was
a package selected in the TX.

Theres no reason we need to hit the registry again
when we've already resolved the package & its providers.
We can cache them to memory to avoid hitting the
registry again for all provider lookups.

The cost of this cache is negligible.

Testing this against installing / removing 139
explicit packages producing 1076 total packages
had the following performance change. Note this
includes blitting and the difference is purely
in resolution time. In `after`, resolution is
almost instant now.

### Before

| command | time | memory |
| - | - | - |
| moss it | 24.6s | 497K   |
| moss rm |  7.8s | 28K    |

### After

| command | time | memory |
| - | - | - |
| moss it | 14.4s | 499K   |
| moss rm |  1.2s | 28K    |

Almost no memory cost and we get near instant resolution
which was a good 6-10s w/out caching.

### DIFF
--- a/moss/src/registry/transaction.rs
+++ b/moss/src/registry/transaction.rs
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
+use std::collections::HashMap;
+
 use dag::Dag;
 use thiserror::Error;
 
@@ -41,6 +43,11 @@ pub struct Transaction<'a> {
 
     /// Dependency lookup strategy
     lookup: Lookup,
+
+    /// Used as a cache to quickly resolve providers for things we've
+    /// already added to the transaction so we don't have to hit the
+    /// registry again
+    selection_providers: HashMap<Provider, package::Id>,
 }
 
 /// Construct a new Transaction wrapped around the underlying [`Registry`].
@@ -53,6 +60,7 @@ pub fn new(registry: &Registry, lookup: Lookup) -> Result<Transaction<'_>, Error
         registry,
         packages: Dag::default(),
         lookup,
+        selection_providers: HashMap::default(),
     })
 }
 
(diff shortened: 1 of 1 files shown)


moss/src/registry/plugin/active.rs
```diff
@@ -24,7 +32,13 @@ impl Eq for Active {}
 impl Active {
     /// Return a new Active plugin for the given state + install database
     pub fn new(state: Option<State>, db: db::meta::Database) -> Self {
-        Self { state, db }
+        let cached_providers = db.all_providers().unwrap_or_default().into_iter().into_group_map();
```
jplatte: Instead of `unwrap_or_default` which swallows any error, how about 

```rust
.unwrap_or_else(|e| {
    tracing::warn!(
        error = &e as &dyn std::error::Error,
        "failed to cache providers",
    );
    Default::default()
})
```

jplatte: Actually, looks like we later _only_ check this cache - so really this should be called `load` [`_into_memory`], right? "cache" seems like the wrong name to me.

tarkah: @jplatte sorry I nuked the 2nd commit which drops all this. Turns out it wasn't needed. The providers cache in the TX itself was sufficient after I added a fix to add the provider to the cache immediately instead of after that package was considered (we pushed it onto a queue).

ermo:
**After v2:**
    
    t sudo moss it -y ptyxis xmlto papers nautilus loupe ldc-devel file-roller dub
    
    245779 entries blitted in 3.67s (67.0k / s)
    
    real 20.49s
    cpu  8.23ms (3.19ms user + 5.03ms sys)
    rss  5.43mb

    t sudo moss rm -y ptyxis xmlto papers nautilus loupe ldc-devel file-roller dub
    
    239411 entries blitted in 3.59s (66.6k / s)
    
    real 33.74s
    cpu  7.05ms (4.49ms user + 2.57ms sys)
    rss  5.43mb

PR #607, opened 2025-10-31
Make it possible to document builtin macro definitions

### LINKED ISSUE #376: boulder: Make boulder's own code (i.e. outside of external .yaml fields) self-documenting re. macros
Make it so that each version of boulder is fully self-documenting in terms of supported macros (actions and definitions), by ensuring that:

- macros have fields where they are defined that contain a non-empty "description" and "example" field (description being empty to should be a build-time error, example should be a warning).
- `boulder recipe macros` can output a human readable and a JSON formatted output of the above.

Some refactoring may be needed, but will ultimately make it so that the canonical documentation for what boulder supports is single-sourced from the boulder version itself.

The relevant source files for this are:
- [src/package.rs](https://github.com/serpent-os/tools/blob/main/boulder/src/package.rs)
- [src/build/job/phase.rs](https://github.com/serpent-os/tools/blob/main/boulder/src/build/job/phase.rs)

Related to #202

jplatte:
WIP (actual docs not written, CLI output not yet updated)
Closes https://github.com/AerynOS/os-tools/issues/376.


ermo:
@jplatte:

Built-in definitions should be shown as `%(name)`, not `%name`. Is this easy to change for you?

For context, think of `%(definition)` as a "constant" and `%action` as a "sub-routine" with a side-effect. This is of course entirely by convention for readability's sake.

jplatte:
Pretty sure this is obsolete now.

ermo:
The branch itself has my documentation work in it. This work can be re-used, hence restoring the branch.

PR #610, opened 2025-11-07
Refactor error report

Preclik02:
Simplified error reporting in moss/src/main.rs.
Replaced the manual sources() function and mutable loop with std::iter::successors to collect the full error chain.
Logging is preserved via tracing::error.

### DIFF
--- a/moss/src/main.rs
+++ b/moss/src/main.rs
@@ -16,22 +16,20 @@ fn main() {
         std::process::exit(1);
     }
 }
-
-/// Display a formatted error message and log the details
-fn report_error(err: &cli::Error) {
-    let chain = collect_sources(err);
-    let message = chain.join(": ");
-    error!(?message, "Command execution failed");
-    println!("{}: {message}", "Error".red());
-}
-
-/// Recursively gather error sources into a string list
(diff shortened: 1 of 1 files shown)


moss/src/main.rs
```diff
@@ -2,36 +2,28 @@
...
-    let mut sources = vec![error.to_string()];
-    let mut source = error.source();
-    while let Some(error) = source.take() {
-        sources.push(error.to_string());
-        source = error.source();
-    }
-    sources
+    // Log with tracing and print to console
+    error!(%chain, "Command execution failed");
+    println!("Error: {chain}");
```
jplatte: Why remove the coloring here?

moss/src/main.rs
```diff
@@ -2,36 +2,28 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use std::error::Error;
+mod cli;
+use crate::cli::Error;
```
jplatte: I prefer the previous import / mod order + qualified use of `cli::Error`.

jplatte:
I'm sorry, it seems I was unclear. I agreed the use of `std::iter::successors` makes sense, I only asked you to revert those changes that I commented on.

PR #612, opened 2025-11-08
boulder: Annotate bsdtar execution errors as such

jplatte:


boulder/src/draft/upstream.rs
```diff
@@ -117,6 +128,10 @@ async fn extract(archive: &Path, destination: &Path) -> Result<(), Error> {
 
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("failed to run `bsdtar`")]
```
ermo, line 131:
```suggestion
    #[error("failed to run `bsdtar`; is it installed and in your PATH?")]
```
Would this work, or is this kind of context non-idiomatic?

ermo (commented):
Cheers for the assist.

boulder/src/draft/upstream.rs
jplatte: I think this is a bit weird to put here, especially because there will be an extra `: {error source}` afterwards.

We can probably do that sort of thing down the line.. somehow 😄

PR #620, opened 2025-11-18
Cargo.toml: Update MSRV to 1.88 for let-chains

ermo:
We have agreed on a MSRV limit of (current stable toolchain version) - 3.


ermo:
> Clippy recognizes that we can use let chains now x) Maybe `cargo clippy --fix` will work. Or do you want me to take care of the warnings?

If you could take care of it, that would be excellent.

tarkah (approved):

PR #622, opened 2025-11-28
boulder: Canonicalize runpath

ReillyBrogan:
There are two changes here:

The first commit fixes rpath handling such that it properly handles rpaths which have `/../` in them. For example `soname(openjdk-17/bin/../lib/libjli.so(x86_64))` which would have resulted in a package that couldn't be installed is now properly handled.

The second commit adds a new key `rundeps-exclude` to the stone yaml and then uses the value of it as a regex to exclude rundeps from being added to packages. This is needed as an escape hatch for instances where the autodep adds a dependency that it shouldn't and it isn't feasible to fix autodep. For example Java appears to use some kind of manual library loading to find `libjvm.so` and then later libraries that depend on it work fine because the manual loading has already happened before the library is loaded. It does not use rpath (for that at least) so autodep can't find the actual path and so adds `libjvm.so` as a dependency. This allows us to work around that.

### DIFF
--- a/boulder/src/package/emit.rs
+++ b/boulder/src/package/emit.rs
@@ -10,6 +10,7 @@ use std::{
 use fs_err::{self as fs, File};
 use itertools::Itertools;
 use moss::{Dependency, Provider, package::Meta};
+use regex::Regex;
 use thiserror::Error;
 use tui::{ProgressBar, ProgressStyle, Styled};
 
@@ -80,6 +81,16 @@ impl<'a> Package<'a> {
                         .iter()
                         .filter_map(|name| Dependency::from_name(name).ok()),
                 )
+                .filter(|dep| {
+                    for exclude_filter in self.definition.run_deps_exclude.iter() {
+                        if let Ok(re) = Regex::new(exclude_filter)
+                            && re.is_match(&dep.to_string())
+                        {
+                            return false;
+                        }
+                    }
+                    true
+                })
                 .collect(),
             providers: self.analysis.providers().cloned().collect(),
             conflicts: self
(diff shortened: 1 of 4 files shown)


boulder/src/package/analysis/handler/elf.rs
```diff
@@ -141,7 +142,7 @@ fn parse_dynamic_section(
             .filter_map(|v| strtab.get(*v).ok())
         {
             for path in rpath.split(':') {
-                let path = path.replace("$ORIGIN", &origin);
+                let path: String = clean(path.replace("$ORIGIN", &origin)).to_string_lossy().to_string();
```
tarkah: Nit: type is inferred here, it doesn't need to be explicitly added

boulder/src/package/emit.rs
```diff
@@ -80,6 +81,18 @@ impl<'a> Package<'a> {
                         .iter()
                         .filter_map(|name| Dependency::from_name(name).ok()),
                 )
+                .filter(|dep| {
+                    let mut allowed = true;
+                    for exclude_filter in self.definition.run_deps_exclude.clone() {
```
tarkah: `iter()` to iterate by reference and remove needless clone

boulder/src/package/emit.rs
```diff
@@ -80,6 +81,18 @@ impl<'a> Package<'a> {
...
+                        if let Ok(re) = Regex::new(&exclude_filter) {
+                            if re.is_match(&dep.to_string()) {
+                                allowed = false;
+                                break;
```
tarkah: Nit: You can `return false` here instead of `break` and remove `allowed` variable, default return `true` as the default returned expression

tarkah (approved):
Code LGTM, thanks! I'd prefer @ermo signs off on the recipe change since we're adding something new.

PR #626, opened 2025-12-12
boulder: Add cleanup flag to build

### LINKED ISSUE #212: boulder: Add command to clean cache
Builders currently don't know how to clean their caches, which means that disks fill up eventually, which in turn causes loss of service and a need for manual intervention.

Implementing a boulder command to clean caches (or prune to X GB via a LRU scheme?) would be useful, particularly if boulder can be set up to do this cleanup itself as a configurable post-build action.

tarkah:
Resolves #212

### DIFF
--- a/boulder/src/build.rs
+++ b/boulder/src/build.rs
@@ -11,6 +11,7 @@ use std::{
 
 use fs_err as fs;
 use itertools::Itertools;
+use moss::repository;
 use nix::{
     sys::signal::Signal,
     unistd::{Pid, getpgrp, setpgid},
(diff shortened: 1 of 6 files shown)


boulder/src/build.rs
```diff
@@ -121,13 +126,29 @@ impl Builder {
...
 
+    pub fn cleanup(&self) -> Result<(), Error> {
+        // Remove rootfs
+        root::remove(self)?;
+
+        // Remove artifacts
+        let _ = fs::remove_dir_all(&self.paths.artefacts().host);
+
+        // Remove build dir
+        let _ = fs::remove_dir_all(&self.paths.build().host);
```
jplatte: Shouldn't we at least log any errors we get from these calls?

boulder/src/build/root.rs
```diff
@@ -52,7 +52,25 @@ pub fn populate(
...
+    util::recreate_dir(&builder.paths.rootfs().host)?;
+
+    Ok(())
+}
+
+pub fn remove(builder: &Builder) -> Result<(), Error> {
+    clean(builder)?;
+
+    // Now we can safely remove the rootfs
+    let _ = fs::remove_dir_all(&builder.paths.rootfs().host);
```
jplatte: Same here.

boulder/src/build.rs
tarkah: I added a proper if exists guard so errors are now relevant

boulder/src/build/root.rs
tarkah: I added a proper if exists guard so errors are now relevant

ermo (commented):
Feature-wise, this makes sense to me.

Will defer to @jplatte re. potential issues that need fixing.

Issue #633, opened 2025-12-17
moss: add the ability to import a system-model


ermo:
Importing a previously exported system-model.kdl file could be very useful for e.g. installer purposes or simply sharing a system-model for reproducibility on other systems.

One suggestion for the way this proposed `moss sync --import <system-model.kdl file>` could work:

- If moss is imperative (= implicit) mode (no system-model file present), we show the diff, ask to sync, and at the end, ask to enable system-model (= write imported model to `/etc/moss/system-model.kdl`), thus reusing existing primitives?
- If moss is in declarative (= explicit) mode (system-model file present), we show the diff, ask to overwrite the imported model to `/etc/system-model.kdl` (do we offer to back up the previous model as an export?), save it on confirm XOR don't save and just do an "ephemeral sync" (= deviate from the existing system-model via the new, imported system-model) otherwise, and then do the actual sync? (= reuse existing primitives?)

PR #637, opened 2025-12-18
Task/moss import export sync

### LINKED ISSUE #632: moss: enable exporting an implicit/imperative system-model to a declarative file
To simplify exporting the system-model to a file, add an output option to moss state export:
```
moss state export -o|--output <filename>
```
This should be equivalent to:
 ```
moss state export > "system-model-fstxn-$(moss state active | gawk '/#/ { gsub(/#/, "", $2); print $2; }').kdl"
```

tarkah:
Resolves #632 & #633

Also updates the output of `sync` to breakout added vs updated vs removed. Where `updated` shows version diffs:

image


ermo (commented):
First two commits look really nice. The third and last commit has a couple of clippy lints and not a lot of documentation?

PR #639, opened 2025-12-19
container: use nc over libc for move_mount for musl support

tarkah:
`nix -> libc` doesn't expose `move_mount` and related things for `musl`. `nc` gives us syscall bindings w/out reliance on `libc`.

The context for this is that we need to be able to build stand-alone, statically compiled boulder binaries against musl for the purpose of convenience of distribution.


ermo (approved):
This looks deceptively simple.

ermo:
> @tarkah said:
> @jplatte `rustix` looks perfect for this. I say we rip out `nix` & now `nc` completely and replace w/ `rustix`

New PR then?

PR #647, opened 2025-12-30
Add AStr type and use it for paths

jplatte:
I should probably not have stayed up long enough to finish this, but I did.

No (heap) profiling has been carried out yet.


tarkah (approved):
Approved pending some memory profiling comparisons

PR #651, opened 2026-01-01
astr: Use ThinArc

jplatte:
Reduces memory usage without impacting performance of my simple vfs-building benchmark.

Before:

```
calls to allocation functions: 3.000.710
peak heap memory consumption: 74,7MB

Benchmark 1: moss -D /mnt/aerynos state build-vfs
  Time (mean ± σ):     411.1 ms ±   2.2 ms    [User: 381.8 ms, System: 27.0 ms]
  Range (min … max):   407.3 ms … 417.8 ms    40 runs
```

After:

```
calls to allocation functions: 3.000.710
peak heap memory consumption: 62,7MB

Benchmark 1: moss -D /mnt/aerynos state build-vfs
  Time (mean ± σ):     412.3 ms ±   2.9 ms    [User: 385.9 ms, System: 26.3 ms]
  Range (min … max):   408.5 ms … 420.9 ms    40 runs
```

### DIFF
--- a/crates/astr/src/lib.rs
+++ b/crates/astr/src/lib.rs
@@ -1,9 +1,12 @@
 use std::{
     borrow::{Borrow, Cow},
     fmt,
+    hash::Hash,
     ops::Deref,
     path::Path,
 };
+
+use triomphe::{Arc, HeaderWithLength};
 
 mod diesel;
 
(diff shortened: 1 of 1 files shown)


crates/astr/src/lib.rs
```diff
@@ -1,22 +1,28 @@
...
-pub struct AStr(triomphe::Arc<str>);
+#[derive(Clone)]
+pub struct AStr(triomphe::ThinArc<(), u8>);
 
 impl AStr {
+    #[inline]
     pub fn as_str(&self) -> &str {
-        &self.0
+        // SAFETY: We only ever store UTF-8,
+        // would use ThinArc<(), str> if possible
```
tarkah, line 24: Well even a String or &str is just a vec / slice of bytes with type safety around it. So no change here either.

tarkah (approved):
Lgtm. Would love some benchmarking numbers posted before we merge.

crates/astr/src/lib.rs
jplatte: Yup, just annoying that we need unsafe here.

jplatte: Yup, just annoying that we need unsafe here.

PR #655, opened 2026-01-02
Store VFS path file_name and parent indices instead of separate strings

jplatte:
`moss state build-vfs` before:

```
calls to allocation functions: 3.000.710
peak heap memory consumption: 62,7MB

Benchmark 1: moss -D /mnt/aerynos state build-vfs
  Time (mean ± σ):     412.3 ms ±   2.9 ms    [User: 385.9 ms, System: 26.3 ms]
  Range (min … max):   408.5 ms … 420.9 ms    40 runs
```

`moss state build-vfs` after:

```
calls to allocation functions: 2.053.987
peak heap memory consumption: 55,8MB

Benchmark 1: moss -D /mnt/aerynos state build-vfs
  Time (mean ± σ):     360.5 ms ±   2.8 ms    [User: 336.1 ms, System: 24.3 ms]
  Range (min … max):   356.1 ms … 370.6 ms    40 runs
```

### DIFF
--- a/crates/vfs/src/tree/builder.rs
+++ b/crates/vfs/src/tree/builder.rs
@@ -5,8 +5,7 @@
 //! Build a vfs tree incrementally
 use std::collections::BTreeMap;
 
-use astr::AStr;
-use elsa::FrozenVec;
+use astr::{AStr, CowAStr};
 
 use crate::path;
 use crate::tree::{Kind, Tree};
@@ -91,7 +90,6 @@ impl<T: BlitFile> TreeBuilder<T> {
             .collect::<BTreeMap<_, _>>();
 
         // build a set of redirects
-        let scratch = FrozenVec::new();
         let mut redirects = BTreeMap::new();
 
         // Resolve symlinks-to-dirs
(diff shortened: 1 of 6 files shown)


crates/vfs/src/tree/builder.rs
```diff
@@ -85,25 +87,27 @@ impl<T: BlitFile> TreeBuilder<T> {
             .iter()
             .filter(|f| f.kind.is_directory())
             .chain(self.implicit_dirs.values())
-            .map(|d| (&d.path, d))
+            .map(|d| (&*d.path, d))
             .collect::<BTreeMap<_, _>>();
 
         // build a set of redirects
+        let scratch = FrozenVec::new();
```
tarkah: Can we cow target instead? I feel it'd be more obvious why it's being used. Took me a second to understand why this was needed?

tarkah (approved):
Lgtm, but one comment if you want to consider it.

crates/vfs/src/tree/builder.rs
tarkah: Or even an option. Does frozen vec have some advantage here?

jplatte: Good point, I'll try that.

PR #662, opened 2026-01-19
stone: Decode unknown variants

tarkah:
Our current decode implementation breaks on unknown variants in our payload data. This makes it impossible to add backwards compatible changes (new meta types) without breaking older moss installs.

This change allows moss to decode unknown variants and effectively ignore them so we can be forwards compatible with future changes that don't require a hard format bump to support (conflicts, replaces, retires, tags, etc) & without having to split those stones into a newer format repository since they can still be understood & used by older installs (just missing non-mandatory functionality).

### DIFF
--- a/crates/stone/src/payload/layout.rs
+++ b/crates/stone/src/payload/layout.rs
@@ -11,8 +11,8 @@ use crate::{ReadExt, WriteExt};
 
 /// Layout entries record their target file type so they can be rebuilt on
 /// the target installation.
-#[repr(u8)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
 pub enum FileType {
     /// Regular file
     Regular = 1,
@@ -34,6 +34,8 @@ pub enum FileType {
 
     /// UNIX Socket
     Socket,
+
+    Unknown = 255,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -47,6 +49,8 @@ pub enum Entry {
     BlockDevice(AStr),
     Fifo(AStr),
     Socket(AStr),
+
+    Unknown(AStr, AStr),
 }
 
 impl Entry {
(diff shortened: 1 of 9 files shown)


crates/stone/src/payload/layout.rs
```diff
@@ -11,11 +11,12 @@ use crate::{ReadExt, WriteExt};
 
 /// Layout entries record their target file type so they can be rebuilt on
 /// the target installation.
-#[repr(u8)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum FileType {
+    Unknown(u8),
```
jplatte: This has the disadvantage of allowing construction of `Unknown(1)` even though `1` is also the value for `Regular` (or w/e). I think we should consider `pub struct FileType(u8);` with associated constants for the well-known values (can also use associated consts in `match` arms).. To reduce diff, you could even `#[allow(non_snake_case)]` or whatever the lint is that fires on `CamelCase` constants.

tarkah: Hmm, I understand the concern but then we lose out on exhaustive match niceties and such.

`Unknown(1)` would have to be constructed by the caller and would really only come into play on the `write` path (since we control the `read` path & our implementation is correct) and in reality `Regular == Unknown(1)` during write anyways :shrug:

tarkah: The alternative is we do not store the unknown tag and Uknown is a concrete tag / `0` we saturate to during decode. We obviously can't roundtrip these but I'm not sure that's really a concern.

I'm looking to rebase our `libstone` branch to this and I think having `Unknown = 0` makes more sense from a C API anyways? 

Theres one enum we shot ourselves in the foot on and its `meta::Dependency` where `0` is `PackageName` so we'd need to reserve `255` for `Unknown` but I think this is fine / just as valid.

tarkah: I will leave the current implementation & saturate all `Unknown` values to `255` / `u16::MAX` for `libstone`

tarkah:
> No strong stance from me on this stuff. I don't love that these enum types are now `u16`s, but it's probably not a big deal.
> 
> I wonder whether it will be possible in the future with Rust's proposed pattern types to do
> 
> ```rust
> #[repr(u8)]
> enum Foo {
>     A = 1,
>     B = 2,
>     Unknown(u8 is 3..),
> }
> ```
> 
> that would be very cool x)

I've added back the `repr` layout as this is needed for `libstone`, which I've rebased that PR to this: https://github.com/AerynOS/os-tools/pull/364

I think both are now ready to merge.

PR #670, opened 2026-01-23
moss: Split cache::Error in two and use snafu for them

jplatte:
Follow-up to #668.

### DIFF
--- a/moss/src/client/cache.rs
+++ b/moss/src/client/cache.rs
@@ -13,8 +13,8 @@ use std::{
 };
 
 use futures_util::StreamExt;
+use snafu::{OptionExt, ResultExt as _, Snafu, ensure};
 use stone::{StoneDecodedPayload, StonePayloadIndexRecord, StoneReadError};
-use thiserror::Error;
 use tokio::io::AsyncWriteExt;
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


moss/src/client/cache.rs
```diff
@@ -322,22 +322,24 @@ pub fn asset_path(installation: &Installation, hash: &str) -> PathBuf {
...
+    // FIXME: Add context, remove context(false)
+    #[snafu(context(false), display("read stone"))]
+    ReadStone { source: StoneReadError },
+    #[snafu(display("parse url {url}"))]
+    ParseUrl { source: url::ParseError, url: String },
+    #[snafu(display("fetch stone {url}"))]
+    FetchStone { source: request::Error, url: String },
+    // FIXME: This could use more detail / splitting up
+    #[snafu(display("io"))]
+    Io { source: io::Error },
```
tarkah: Can we split this up now into more helpful operation related variants? This PR can then serve as more of a reference for other future work.

moss/src/client/cache.rs
```diff
@@ -87,19 +87,20 @@ pub async fn fetch(
 ) -> Result<Download, Error> {
     use fs_err::tokio::{self as fs, File};
 
-    let parse_url = |s: &str| s.parse::<Url>().map_err(|err| Error::ParseUrl(err, s.to_owned()));
+    let parse_url = |url: &str| url.parse::<Url>().context(ParseUrlSnafu { url });
 
-    let url = parse_url(meta.uri.as_deref().ok_or(Error::MissingUri)?)?;
+    let url_s = meta.uri.as_deref().ok_or(Error::MissingUri)?;
+    let url = parse_url(url_s)?;
```
tarkah: We don't need `parse_url` closure since you already bind an intermediary `url_s` variable we can use to inline `url.parse::<Url>().context(ParseUrlSnafu { url })`?

moss/src/client/cache.rs
```diff
@@ -322,22 +317,28 @@ pub fn asset_path(installation: &Installation, hash: &str) -> PathBuf {
...
+    #[snafu(display("missing hash"))]
+    MissingHash,
+    #[snafu(display("malformed hash `{hash}`"))]
+    MalformedHash { hash: String },
+    #[snafu(display("missing URL"))]
+    MissingUrl,
+    #[snafu(display("invalid URL `{url}`"))]
+    InvalidUrl { source: url::ParseError, url: Box<str> },
+    #[snafu(transparent, context(false))]
+    Request { source: request::Error },
```
tarkah, line 341: Doesn't this now have less context than before? Or is `Uri` always displayed via reqwest error?

jplatte: Ah, should have left a comment about this. I figured users would care what package failed to download, the URL moss was trying to download from seems like noise almost all of the time. And it's pulled straight from the package metadata, where it should be very easy to grab it again in case this error ever needs in-depth troubleshooting.

Issue #488, opened 2025-05-17
moss: Add `fetch`, `cache seed` and `cache export` commands


ermo:
> **ermo**
> How much work would it be to do a thing that could fetch all the known .stones from a repo? Would that be a moss command?
> i.e. `moss fetch <uri>`? Or would it be a standalone tool?
> Context: I would like it to be easier to seed a vessel import.
> so having a way to say "get me these packages from this index and save them to this folder" would be neat.
> I know we have a hacky sqlite based way to do it ofc
> **tarkah**
> Probably a builtin to moss would be easiest
> **ermo**
> I mean, suppose we added a fetch command that could then either take a list of pkgs to fetch XOR a "special" keyword such as "all"?
> we still have the problem that moss does not yet create an easily readable list of fetched stones
> so maybe we could add a separate dir that only contains a full list of plain symlinks to where the .stones are fetched?
> (that we can "refresh" by walking the download / asset cache)
> I hope what I am getting at is making sense. If I am misremembering how it works, do correct me.
> **tarkah**
> This command wouldn't work on the .moss root, it should just expect an output dir and use the package name, version, relno to save each stone
> Unless you want it working on the moss root
> That'd be like a cache command
> Vs fetch
> **ermo**
> I think we should have both honestly
> **tarkah**
> Pre-cache
> Yup
> **ermo**
> Because pre-cache is super nice
> so this would be more a sort of "export" command
> like, you precache all current index stones. Then you export those via copy_or_hardlink?
> seems like a faster way of doing it...
> **tarkah**
> Yeah could just do it that way
> **ermo**
> i.e. "export either the named or all the .stones belonging to the current index"
> **tarkah**
> Then it only works with a valid moss root available vs standalone
> **ermo**
> which means "current index" is "whatever was sync'd" for context
> which is going to be pretty handy for versioned repos.
> where you can sync to a repo version
> (currently it'd just be "whatever is now the current stone.index")
>
>> **tarkah**
>> Then it only works with a valid moss root available vs standalone
>
> yeah, but don't we get a hella lot for free if we use a moss root?
> i.e. code we don't have to write?
> the alternative is this is a feature in vessel via cli
> where you point to a stone.index and go "fetch these as your pre-cache/seed"
> **tarkah**
> Yeah but we can also refactor things to work with and without moss root
> Without always just saves it as human readable to a dir
> **ermo**
>
>> **tarkah**
>> Without always just saves it as human readable to a dir
>
[...]

PR #674, opened 2026-01-27
(feat) moss: Implement support for fetching packages

joebonrichie:
image

Part of #488

Depends on #673, #675

There is a bit of code duplication here with `client::fetch_packages()` however, we explicitly do not wish to unpack packages or add them to the install db only fetch them. If a blocker, I can refactor cache_packages so it can handle both use cases.

### DIFF
--- a/moss/src/cli/fetch.rs
+++ b/moss/src/cli/fetch.rs
@@ -2,51 +2,46 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
-use clap::{ArgMatches, Command, arg, value_parser};
-
+use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
 use moss::{Installation, client::Client, environment};
 use tracing::instrument;
 
 pub use moss::client::Error;
 
-pub fn command() -> Command {
-    Command::new("fetch")
-        .visible_alias("fe")
-        .about("Fetch package(s)")
-        .long_about("Fetch package stone(s) by name")
-        .arg(arg!(<NAME> ... "packages to fetch").value_parser(clap::value_parser!(String)))
-        .arg(
-            arg!(-o --"output-dir" <OUTPUT_DIR> "directory to write the fetched stone(s) to (defaults to OUTPUT_DIR)")
-                .value_parser(value_parser!(PathBuf)),
-        )
+pub fn command() -> clap::Command {
+    Command::command()
+}
+
+#[derive(Debug, Parser)]
+#[command(
+    name = "fetch",
+    visible_alias = "fe",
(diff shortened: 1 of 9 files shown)


moss/src/cli/fetch.rs
```diff
@@ -0,0 +1,53 @@
...
+use clap::{ArgMatches, Command, arg, value_parser};
+
+use moss::{Installation, client::Client, environment};
+use tracing::instrument;
+
+pub use moss::client::Error;
+
+pub fn command() -> Command {
+    Command::new("fetch")
+        .visible_alias("fe")
```
tarkah: Shortcut aliases also have to be added to a replacement array in the cli root module to take effect since clap doesn't support this ootb

moss/src/cli/fetch.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    if let Some(dir) = args.get_one::<PathBuf>("output-dir")
+        && !dir.exists()
+    {
+        std::fs::create_dir_all(dir)?;
+    }
+
+    let output_dir = match args.get_one::<PathBuf>("output-dir") {
+        Some(dir) => &dir.canonicalize()?,
+        None => &PathBuf::from(".").canonicalize()?,
```
tarkah: Path::new from a static str, this will avoid a needless allocation

moss/src/client/fetch.rs
```diff
@@ -0,0 +1,254 @@
...
+        true
+    } else {
+        Confirm::with_theme(&ColorfulTheme::default())
+            .with_prompt(" Do you wish to continue? ")
+            .default(false)
+            .interact()?
+    };
+    if !result {
+        return Err(Error::Cancelled);
+    }
```
tarkah: This operation doesn't mutate state, do we really need a confirmation?

moss/src/client/fetch.rs
```diff
@@ -0,0 +1,254 @@
...
+                .unwrap();
+
+            let dest_path = output_dir.join(file_name);
+
+            let multi_progress = multi_progress.clone();
+            let total_progress = total_progress.clone();
+
+            runtime::unblock(move || {
+                std::fs::hard_link(download.path(), &dest_path).unwrap();
+
```
tarkah: I think we have a hard link or copy utility function we can use. Or maybe we need to create a moss::util module and move it there.

We will face lots of crosslink device errors when people try to output to a different mount, like /tmp

joebonrichie: Ah good point, i'll move it to `moss::util`. I also see that it contains an `enumerate_files()` function which can probably replace the indexing specific enumerate_files, may also be a few more cleanup opportunities.

joebonrichie: #675

PR #681, opened 2026-02-02
moss: Improve extraction of individual stones

joebonrichie:
- Refactor the blitting code such that it doesn't require a moss client

- Add an extraction method in moss client to extract individual .stone(s) using the blitting code

- Move over moss/cli/extract to use the new extraction method resolving a couple of fixmes in the process

### DIFF
--- a/moss/src/client/extract.rs
+++ b/moss/src/client/extract.rs
@@ -32,83 +32,84 @@ pub fn extract(stones: Vec<PathBuf>) -> Result<(), Error> {
             .find_map(StoneDecodedPayload::meta)
             .ok_or(Error::MissingMeta)?;
 
-        if let Some(layouts) = layouts {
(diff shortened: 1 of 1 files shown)


moss/src/client/extract.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        let payloads = reader.payloads()?.collect::<Result<Vec<_>, _>>()?;
+        let content = payloads.iter().find_map(StoneDecodedPayload::content);
+        let layouts = payloads.iter().find_map(StoneDecodedPayload::layout);
+        let meta = payloads
+            .iter()
+            .find_map(StoneDecodedPayload::meta)
+            .ok_or(Error::MissingMeta)?;
+
+        if let Some(layouts) = layouts {
```
tarkah:
```rs
let Some(layouts) = layouts else {
  println(..);
  continue;
}
```

Removes some indentation

PR #687, opened 2026-02-05
[WiP] moss: switch cli to use clap derive structs

### LINKED ISSUE #196: moss: Discuss a CLI
We need a sane CLI for moss. Let's talk about subcommands and flags.

NB: Use this [Tree generator](https://tree.nathanfriend.com/?s=(%27options!(%27fancy!true~fullPath!false~trailHgSlash!false~rootDot!false)~source!(%27source!%27moss38yEadB%7BraWAdd5a%20new%208Jto_E*%3Curl%3EEliZMrlWLiZ58ie5H24ErX6%7BrrWRXe5a%208Jfrom_37EadB%7BpaWAdB7%7BsNto2underlyj4%20fromI%20fil6path%20oQbJnam6H2repoFE*--reHZallEliZMpl9HZalleB75accordjto2resolution%20of24E*explicitMple9explicitlJHZalleB75H24.EIY%7Bpla9aY75given2az68ie5H24.ErX6%7BprmWRXe57%7BsNfrom24F3searchMsrWSearche5%25r5%7BHcludj7sNbJnam6anBsummary.E%24%20...E-p%2C--%255%24*Syntax%20sugaQfoQusjnativ6%22foo%7Bbar%7D%22%20%25Qsearches3CEazat6Azate2given%20CEaz6ShowIze*%20CsEliZ*LiZIYCEprun6Prun6all%20but2laZTk%20C5%7Bk%3D10%20is2default%7DE-k%2C--keep%20Keep%20thi5manJCsErX6RX6arbitrarJCs.%20Support5sHgl6C5anBHclusiv6range5a-b3sync*Sync%2Bto2current%204E-u%2C--update*Updates28ie5defHeBH24%20prioQto%20sync%22Hg%5Cn%5Cn%20%23%20Global%20options3TD%2CT-directory*Override%2Broot.3Th%2CT-help.***Show5helpInBquits.%20ValiBfoQsubcommanBhelp5too.3T-%26****PrHt5mos5%26InBquits.%27)~%26!%271%27)*%20%202%20th63%5Cn*4syZem-model5s%206e%207package8repositor9NLiZIll%20Bd%20CZateE3*F.E*%3C7name%3E%20...HinI%20aJy%20M%20%7BN%7D%20Qr%20T%20-W%7D*XemovYvailabl6Zst_24E*%3Cname%3EjHg%20zctiv%24%3Ckeyword%3E%25provide%26version%2Bs2syZem%20%01%2B%26%25%24zj_ZYXWTQNMJIHFECB98765432*) for changing the layout below.

```text
moss
├── repository
│   ├── add (ra)  Adds a new repository to the system-model
│   │   ├── <name>
│   │   └── <url>
│   ├── list (rl)  Lists repositories in the system-model
│   └── remove (rr)  Removes a repository from the system-model
│       └── <name>
├── package
│   ├── add (pa)  Add package(s) to the underlying system-model from a file path or by name in the repo.
│   │   ├── <packagename> ...
│   │   └── --reinstall
│   ├── list (pl) List all installed packages according to the resolution of the system-model
│   │   ├── explicit (ple) List all explicitly installed packages in the system-model.
│   │   └── available (pla) List all available packages given the active repositories in the system-model.
│   └── remove (prm)  Removes package(s) from the system-model.
│       └── <packagename> ...
├── search (sr)  Searches providers (including packages) by name and summary.
│   ├── <keyword> ...
│   └── -p,--provides <keyword>  Syntax sugar for using native 'foo(bar)' provider searches
├── state
│   ├── activate Activate the given state
│   ├── active Show active   states
│   ├── list  List available state
│   ├── prune Prune all but the last -k states (k=10 is the default)
│   ├── -k,--keep Keep this many states
│   └── remove Remove arbitrary states. Supports single states and inclusive ranges a-b
├── sync  Syncs the system to the current system-model
│   └── -u,--update  Updates the repositories defined in the system-model prior to sync'ing
└── # Global options
    ├── -D, --directory  Overrides the system root.
    ├── -h, --help.      Shows help and quits. Valid for subcommand helps too.
    └── --version        Prints moss version and quits.
```

### LINKED ISSUE #611: moss: Unexpected command line parsing behaviour
As below:

```
barrylhm@irregularpc ~> moss search ar
error: unexpected argument 'add' found

Usage: moss search [OPTIONS] <KEYWORD>

For more information, try '--help'.
```
This behavior is same as
`moss search repo add` so probably moss expanded argument `ar` as a command.
This is not a severe problem but kinda annoying

ermo:
This PR migrates the CLI from function-based Clap declarations to clap_derive. clap_derive allows a declarative-like definition of the CLI command tree, so it's easier to maintain and extend.

Inside `moss/src/cli` we'll have  a root `mod.rs` file and one module per each subcommand. Modules may be split into other submodules when the logic becomes complex and/or multiple tests are defined (e.g. for the `package` subcommand).\
Each module will have a `Command` struct with `handle()` as its only public method. The signature of each `handle()` is not identical for all `Command`s, but depends on the logic it performs.\
Public structs/enum/functions will be at the top of each source file, private ones will be right below.

This PR fixes #196, but that issue plan functionalities that the moss backend does not yet support, and that's the reason why you'll see a bunch of `unimplemented!()` calls across the PR.\
It also fixes #611.

Another thing I wanted to make consistent is the use verbs in their imperative conjugation instead of third-person, as these are *commands*. I also chose not to end phrases with a dot, like most programs out there.

I think the code could be cleaned a little (e.g. there are some Vec conversions I'm not sure they would be needed with more generic code), but that's out of the scope of this PR. I'm trying to introduce no functional and API changes here, except the parsing of CLI arguments: where possible, I'm leveraging Clap's automatic call of `TryFrom` to sanitize inputs, particularly for providers where I'm now passing the `Provider` struct.

TODO: print final CLI tree.


ermo:
The `moss package inspect` alias (`pin`) might need to change, as the plan is to add an actual `moss package pin` command that saves a package in a default highest-priority cobble repo and keeps it there until the user removes it (which will typically happen when it causes the system-model to not be able to resolve).

I think `pit` might work, as in first-and-last-letter-of-`inspect` ?

ermo:
For ease of use, we *might* consider allowing `pkg` instead of `package`, given we already allow `repo` instead of `repository`?

What is your thinking on potentially allowing both `package` and `pkg` as well as `repo` and  `repository`, while only showing one of each in help?

livingsilver94:
> The `moss package inspect` alias (`pin`) might need to change, as the plan is to add an actual `moss package pin` command that saves a package in a default highest-priority cobble repo and keeps it there until the user removes it (which will typically happen when it causes the system-model to not be able to resolve).

Not a problem, I'll do that. But I want to highlight that a `pin` alias and a `pin` sub-subcommand wouldn't conflict with each other, assuming my algorithm for expanding aliases will be accepted. Granted, that'd be confusing for users.

> For ease of use, we might consider allowing pkg instead of package, given we already allow repo instead of repository?

**Yes please** :smile: 

> What is your thinking on potentially allowing both package and pkg as well as repo and repository, while only showing one of each in help?

I think I could just add a non-visible alias from clap_derive for that but to be honest, I don't think users will type unshortented "repository" or "package" unless forced to, as it is currently the case with "package".

ermo:
If we make moss help show `pkg` and `repo` (while also allowing `package` and `repository` respectively) I think that'd be the nicest?

ermo:
@livingsilver94 I can't get `pkg add/pad` or `pkg info/pif` to work with `man-pages` when using: 

```
moss version v0.26.8 (Git ref 9b31d327c04aed303ecf686dd13e2b91de8ddd4e-dirty) (Built at 2026-09-02T11:13:35+00:00)
```

Examples:

image

livingsilver94:
> @livingsilver94 I can't get `pkg add/pad` or `pkg info/pif` to work with `man-pages`

Fixed.

ermo:
Could you rebase this onto the fnmatch PR?

PR #701, opened 2026-02-16
Execute triggers in parallel batches

joebonrichie:


crates/dag/src/lib.rs
```diff
@@ -103,6 +104,37 @@ where
...
+
+        while g.node_count() > 0 {
+            let mut sources: Vec<_> = g.externals(Direction::Incoming).collect();
+            if sources.is_empty() && g.node_count() > 0 {
+                // Cycle detected.
+                break;
+            }
+
+            // Deterministic output
+            sources.sort_by_key(|a| &g[*a]);
```
tarkah: Is it not already deterministic based off node index / order? Or is `externals` non-deterministic somehow?

crates/dag/src/lib.rs
```diff
@@ -103,6 +104,37 @@ where
...
+
+            let batch_nodes: Vec<_> = sources.iter().map(|&i| g[i].clone()).collect();
+            batches.push(batch_nodes);
+
+            // Reverse index before removing nodes to avoid graph invalidation (dupes in batches)
+            sources.sort_by_key(|&idx| std::cmp::Reverse(idx.index()));
+
```
tarkah, line 126: Can you explain this further, not sure I understand how removing these nodes in any order causes graph invalidation. Nodes should be removable in any order?

moss/src/client/postblit.rs
```diff
@@ -145,57 +153,131 @@ pub(super) fn triggers<'a>(
...
+        triggers.iter().try_for_each(|batch| {
+            batch.par_iter().try_for_each(|trigger| {
+                let completed = counter.fetch_add(1, Ordering::Relaxed);
+                (on_progress)(Progress {
+                    completed: completed as u64,
+                    item: match trigger.handler() {
+                        Handler::Run { run, .. } => run,
+                        Handler::Delete { .. } => "delete operation",
+                    },
+                });
```
tarkah, line 265: Shouldn't this be invoked _after_ executing the trigger?

moss/src/client/postblit.rs
```diff
@@ -145,57 +153,131 @@ pub(super) fn triggers<'a>(
...
+    triggers: &[Vec<TriggerRunner>],
+    on_progress: impl Fn(Progress<'_>) + Send + Sync,
+) -> Result<(), Error> {
+    match scope {
+        TriggerScope::Transaction(install, scope) => {
+            execute_transaction_triggers(install, scope, triggers, &on_progress)?;
+        }
+        TriggerScope::System(install, scope) => {
+            execute_system_triggers(install, scope, triggers, &on_progress)?;
+        }
```
tarkah: The `execute_` functions seem to rebuild `TriggerScope` internally... this is very awkward. You can instead do `scope @ TriggerScope::Transaction(install, _)` in the match statement to pass `scope: TriggerScope` into the called functions and not have to reconstruct it.

moss/src/client/postblit.rs
```diff
@@ -145,57 +153,131 @@ pub(super) fn triggers<'a>(
...
-                        .bind_rw(self.scope.guest_path("usr"), "/usr")
-                        .work_dir("/");
-
-                    Ok(isolation.run(|| execute_trigger_directly(&self.trigger))?)
-                }
-            }
-        }
+    pub fn trigger(&self) -> &CompiledHandler {
+        &self.trigger
     }
```
tarkah: Useless getter when the field can be used directly. Its not even referenced outside this module so its fine the field is private.

crates/dag/src/lib.rs
joebonrichie: Honestly, i can't tell if `externals()` is deterministic or not

crates/dag/src/lib.rs
joebonrichie: petgraph seems to move nodes around once you remove one, invalidating the just collected externals. Reversing means you remove by highest indices first keeping the externals valid (at least from what i can tell)

Cloning the graph into a [StableGraph](https://docs.rs/petgraph/latest/petgraph/stable_graph/struct.StableGraph.html) instead should also work without requiring the reverse trick

tarkah: Makes sense 👍

tarkah (approved):
LGTM! Thanks @joebonrichie

joebonrichie:
> How do we know that two seemingly unrelated triggers won't each write to the same file, meaning the trigger outcome is not deterministic (because the output might be different than if we knew that one would always come after the other)?

In a nutshell, we don't. We have no mechanism for recording what files each trigger writes to

PR #739, opened 2026-03-19
moss/search: Improve search and `provides` commands

otherJL0:
I've worked on some of the suggestions from https://github.com/AerynOS/os-tools/issues/729 plus a few improvements to search:

* `moss search` supports searching on the provider syntax, ie `moss search 'sysbinary(zramctl)'`
* `moss search` groups matches based on matches in package name or package summary
* `moss search` highlights matched substrings in package name or package summary

Screenshot_20260329_232406

Changes not yet implemented include grouping matches based on provider type and searching based on file path.

### DIFF
--- a/moss/src/cli/search.rs
+++ b/moss/src/cli/search.rs
@@ -38,7 +38,22 @@ pub fn command() -> Command {
             Arg::new(FLAG_PROVIDES)
                 .short('p')
                 .long("provides")
-                .num_args(0)
+                .num_args(0..=1)
+                .require_equals(true)
+                .default_missing_value("binaries")
+                .value_parser([
+                    "binaries",
+                    "library",
+                    "name",
+                    "soname",
+                    "pkgconfig",
+                    "interpreter",
+                    "cmake",
+                    "python",
+                    "binary",
+                    "sysbinary",
+                    "pkgconfig32",
+                ])
                 .help("Search for packages providing a binary"),
         )
 }
@@ -46,7 +61,8 @@ pub fn command() -> Command {
 pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error> {
     let keyword = args.get_one::<String>(ARG_KEYWORD).unwrap();
     let only_installed = args.get_flag(FLAG_INSTALLED);
-    let provides = args.get_flag(FLAG_PROVIDES);
+    // let provides = args.get_flag(FLAG_PROVIDES);
+    let provides = args.get_one::<String>(FLAG_PROVIDES).map(|s| s.as_str());
 
     let client = Client::new(environment::NAME, installation)?;
     let flags = if only_installed {
@@ -55,8 +71,8 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         package::Flags::new().with_available()
     };
 
-    let output = if provides {
-        provides_package(&client, flags, keyword)
+    let output = if let Some(provider) = provides {
+        provides_package(&client, flags, provider, keyword)
     } else {
         search_packages(&client, flags, keyword)
     };
@@ -90,7 +106,7 @@ fn search_packages(client: &Client, flags: package::Flags, keyword: &str) -> Vec
     }
 }
 
-fn search_providing_packages_by_kind(
+fn provides_package_by_kind(
     client: &Client,
     flags: package::Flags,
     name: &str,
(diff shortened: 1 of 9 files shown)


moss/src/cli/search.rs
```diff
@@ -117,12 +208,127 @@ impl ColumnDisplay for Output {
...
+    use std::path::Path;
+    use std::sync::LazyLock;
+
+    use super::*;
+
+    static TEST_CLIENT: LazyLock<Option<Client>> = LazyLock::new(|| {
+        let root = Path::new(env!("CARGO_MANIFEST_DIR")).join("../aosroot");
+        let installation = Installation::open(root, None).ok()?;
+        Client::new("TEST", installation).ok()
+    });
```
tarkah: We shouldn't have tests rely on the availability of a non-deterministic ignored local directory.

We should have some `Client::mocked` method or something that is only exposed for `#[cfg(test)` target and uses a tempdir backed Installation & allows us to define a registry w/ `registry::plugin::Test`  / test plugin where we can load the explicit packages we want to test this logic against.

moss/src/cli/search.rs
```diff
@@ -56,59 +81,125 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
+fn provides_package(
+    client: &Client,
+    flags: package::Flags,
+    provider: &str,
+    name: &str,
+) -> Result<BTreeMap<MatchKind, Vec<Output>>, Error> {
+    let mut result: BTreeMap<MatchKind, Vec<Output>> = BTreeMap::new();
+    let packages = match provider.parse::<dependency::Kind>() {
+        Ok(kind) => provides_package_by_kind(client, flags, name, kind),
+        Err(_) => match provider {
```
tarkah: Err(_) branch should be unreachable here since we use `value_parser` with all valid options. Regardless, we shouldn't fall through and should bail if an invalid `provides -> provide kind` is parsed here.

moss/src/cli/search.rs
```diff
@@ -56,59 +81,125 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
+                output_kind.entry(match_kind).or_default().push(Output {
+                    name: pkg.meta.name,
+                    summary: pkg.meta.summary,
+                    search_match: Some(keyword.to_owned()),
+                });
+            }
+            Ok(output_kind)
+        }
+        _ => {
+            output_kind.insert(
```
tarkah: I don't think we should deal w/ `Provider` at all here and should just delegate directly to keyword searches here.

If input has `(` and parses into a valid `Provider` OR `--provides` is supplied & parses into `Provider`, then we can branch into the `provides_package` function which will take `Provider` instead of the deconstructed pieces.

I think this will streamline both these functions a lot & not leak any `provides` logic into the `search by keyword` path.

moss/src/cli/search.rs
otherJL0: Sorry this one took me a while. I'm still pretty new to Rust and had some trouble figuring out why my tests were complaining that my mocked methods did not exist. 

My current understanding is that `moss` is both a library crate and a binary crate with the library crate as a dependency. The methods gated behind the `#[cfg(test)]` feature are scoped within the crate, so when I build out my test suite which is part of the binary crate, I'm trying to call mocked methods from the library crate that don't exist because they weren't compiled. So the fix as I understand it is to gate these methods behind a custom feature (ie "testing") which means the library crate is compiled with the gated methods which can be called from the binary crate.

Let me know if I got any of this wrong so far. I'm unblocked now and I should be able to include the test suite pretty soon.

ermo: @otherJL0 If you could join https://aerynos.zulipchat.com and sync up with tarkah, that would probably be wise?

tarkah:
Looks good! Nice update to make testing deterministic using a new Client::mocked helper w/ in memory sqlite dbs. 

One UI feedback. I'm not sure we need the split between name & summary matches w/ grouped output. I feel the highlighting gets the point across well enough. An example below where I removed those groupings vs w/ them

| With grouping headers | Sorted by grouping | No headers / sort |
| - | - | - |
| image | image | image |

I personally find the latter 2 fine, where we can still de-prioritize the summary matches by sorting them after the name matches but not printing the header.

This can be paired with a `--name-only` or `--field=name OR --field=summary` type params in the future if someone really wants to only see matches from one. Omitting header helps standardize this display.

ermo:
@otherJL0 Mind rebasing this so I can begin testing? We've managed to finish up all the other open PRs, so it'd be nice to be able to get this one over the finish line soon; perhaps even before the end of the month.

tarkah:
Please squash, history is pretty crazy now

ermo:
@otherJL0 Do you have an account on our https://aerynos.zulipchat.com instance yet...?

ermo:
> Please squash, history is pretty crazy now

@tarkah Could you give some guidance on how much squash is enough here?

otherJL0:
Hey sorry I rebased off my fork which is stale, working on fixing

CookieSource:
this is closed?  a new one will be opened?

otherJL0:
> this is closed? a new one will be opened?

@tarkah @ermo @CookieSource I think I need to brush up on my git skills, I dug myself into a weird git state but I got it fixed now 😅. This PR should be good to go!

ermo:
@otherJL0 Does this need to be _\*checks notes\*_ 42 commits? Could some of them be squashed together?

Also, `just test` appears to fail?

otherJL0:
@ermo Good callout, I've squashed some of the commits to a manageable amount. I've fixed the `just test` issue, I forgot to add the `--features moss/testing` flag like I did in the ci.yaml file earlier.

ermo:
In the context of #729 :

Is it feasible to have and `moss search -p /usr/bin/bash` be turned into `moss search -p 'binary(bash)'` in this PR?

And similarly, is it feasible to have `moss search -p /usr/lib/libc.so.6` be turned into `moss search -p 'soname(libc.so.6)'`?

Or is this going to have to be a follow-up PR?

tarkah:
Let's tackle these in a follow up so we can get this work merged

tarkah (approved):
LGTM! Thanks for this, sorry it took forever to land.

PR #746, opened 2026-03-24
Introduce the gitwrap crate and refactor how we deal with Upstreams in Boulder

livingsilver94:
This PR introduces the `gitwrap` crate to handle Git repositories from a centralized API. This crate calls the `git` executable under the hood. It's not ideal, but neither libgit2 nor gitoxide support all operations we need to perform (as of April 2026), particularly worktrees.

It also refactors the code for storing and *sharing*[1] plain and Git upstreams.\
With regards to storing, we are now storing repositories in Mirror mode, under directories with unique names composed of the host they were downloaded from, and the very repository's name, e.g. `invent.kde.org_qt.qt.qtdeclarative.git`.
With regards to sharing, we clone the stored repository into the destination directory, also copying the Git remotes from the original ("centralized") repository. This ensures Git sub-modules with relative paths are resolved against the original remote and not against our local mirror. This ensures the removal of our `staging` option, that was developed to work around Git submodules with relative paths.

This PR also restores the `clonedir` option, that I initially removed under assumptions that not shared by the rest of the Team.

[1] We call *sharing* the act of copying an upstream to a build directory when building a recipe. We try to perform such copy as efficiently as possible, e.g. by using hard links.

### Test plan

Recipes I successfully built:
- g/gexiv2/stone.yaml
- f/font-monaspace/stone.yaml
- n/nano/stone.yaml
- q/qt5-declarative/stone.yaml
 
I built all recipes twice, to check that the upstream is downloaded only once.

I also made sure that `%break_continue`, `%break_exit` and plain `boulder chroot` work.

---

This is a refactor. Code inside `boulder` should be reviewed per file, instead of per diff.

### DIFF
--- a/boulder/src/build.rs
+++ b/boulder/src/build.rs
@@ -120,7 +120,7 @@ impl Builder {
         timing: &mut Timing,
         initialize_timer: timing::Timer,
         update_repos: bool,
-    ) -> Result<Vec<upstream::Shared>, Error> {
+    ) -> Result<Vec<upstream::Stored>, Error> {
         // Recreate artifacts
         util::recreate_dir(&self.paths.artefacts().host).map_err(Error::RecreateArtefactsDir)?;
 
@@ -133,7 +133,8 @@ impl Builder {
         let timer = timing.begin(timing::Kind::Fetch);
 
         // Sync (fetch & share) upstreams to rootfs
-        let shared = upstream::sync(
+        let stored = upstream::sync(
+            &self.recipe,
             &self.upstreams,
             &self.paths.upstreams().host,
             &self.paths.guest_host_path(&self.paths.upstreams()),
@@ -141,7 +142,7 @@ impl Builder {
 
         timing.finish(timer);
 
-        Ok(shared)
+        Ok(stored)
     }
 
     pub fn cleanup(&self) -> Result<(), Error> {
--- a/boulder/src/build.rs
+++ b/boulder/src/build.rs
@@ -88,7 +88,7 @@ impl Builder {
             })
             .collect::<Result<Vec<_>, job::Error>>()?;
 
-        let upstreams = upstream::parse(&recipe)?;
+        let upstreams = upstream::parse_recipe(&recipe)?;
 
         let profiles = profile::Manager::new(&env);
         let repos = profiles.repositories(&profile)?.clone();
@@ -115,7 +115,12 @@ impl Builder {
         })
     }
 
-    pub fn setup(&self, timing: &mut Timing, initialize_timer: timing::Timer, update_repos: bool) -> Result<(), Error> {
+    pub fn setup(
+        &self,
+        timing: &mut Timing,
+        initialize_timer: timing::Timer,
+        update_repos: bool,
+    ) -> Result<Vec<upstream::Shared>, Error> {
         // Recreate artifacts
         util::recreate_dir(&self.paths.artefacts().host).map_err(Error::RecreateArtefactsDir)?;
 
@@ -128,11 +133,15 @@ impl Builder {
         let timer = timing.begin(timing::Kind::Fetch);
 
         // Sync (fetch & share) upstreams to rootfs
-        upstream::sync(&self.recipe, &self.paths, &self.upstreams)?;
+        let shared = upstream::sync(
+            &self.upstreams,
+            &self.paths.upstreams().host,
+            &self.paths.guest_host_path(&self.paths.upstreams()),
+        )?;
 
         timing.finish(timer);
 
-        Ok(())
+        Ok(shared)
     }
 
     pub fn cleanup(&self) -> Result<(), Error> {
@@ -150,7 +159,7 @@ impl Builder {
         }
 
         // Remove downloaded upstreams
-        upstream::remove(&self.paths, &self.upstreams)?;
+        upstream::remove(&self.paths.upstreams().host, &self.upstreams)?;
 
         // Prune moss cache, retaining stones from the repos defined
         // by our boulder profile
--- a/boulder/src/cli/build.rs
+++ b/boulder/src/cli/build.rs
@@ -105,7 +105,7 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     );
     println!("boulder {}", tools_buildinfo::get_simple_version());
     println!("└─ building {pkg_name}-{build_release}\n");
-    let shared_upstreams = builder.setup(&mut timing, timer, update)?;
+    builder.setup(&mut timing, timer, update)?;
 
     let paths = &builder.paths;
     let networking = builder.recipe.parsed.options.networking;
@@ -149,9 +149,6 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     package::sync_artefacts(paths).map_err(Error::SyncArtefacts)?;
 
     if cleanup {
-        for share in shared_upstreams {
-            share.remove().map_err(|e| Error::Cleanup(build::Error::from(e)))?;
-        }
         builder.cleanup().map_err(Error::Cleanup)?;
     }
 
--- a/boulder/src/cli/build.rs
+++ b/boulder/src/cli/build.rs
@@ -105,7 +105,7 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     );
     println!("boulder {}", tools_buildinfo::get_simple_version());
     println!("└─ building {pkg_name}-{build_release}\n");
-    builder.setup(&mut timing, timer, update)?;
+    let shared_upstreams = builder.setup(&mut timing, timer, update)?;
 
     let paths = &builder.paths;
     let networking = builder.recipe.parsed.options.networking;
@@ -149,6 +149,9 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     package::sync_artefacts(paths).map_err(Error::SyncArtefacts)?;
 
     if cleanup {
+        for share in shared_upstreams {
+            share.remove().map_err(|e| Error::Cleanup(build::Error::from(e)))?;
+        }
         builder.cleanup().map_err(Error::Cleanup)?;
     }
 
--- a/boulder/src/cli/build.rs
+++ b/boulder/src/cli/build.rs
@@ -177,19 +177,14 @@ fn verify_versions_match(builder: &Builder) -> Result<(), Error> {
 
     // We won't attempt to parse git upstreams for now
     match &first_upstream.props {
-        stone_recipe::upstream::Props::Git { git_ref, staging: _ } => {
+        stone_recipe::upstream::Props::Git { git_ref, .. } => {
             // If we have a git ref, we have a git upstream and version parsing
             // will not work.
             if !git_ref.is_empty() {
                 return Ok(());
             }
         }
-        stone_recipe::upstream::Props::Plain {
-            hash: _,
-            rename: _,
-            strip_dirs: _,
-            unpack: _,
-        } => {}
+        stone_recipe::upstream::Props::Plain { .. } => {}
     }
 
     let ver_ext = VersionExtractor::new();
--- a/boulder/src/upstream.rs
+++ b/boulder/src/upstream.rs
@@ -162,7 +162,7 @@ pub fn sync(upstreams: &[Upstream], storage_dir: &Path, share_dir: &Path) -> Res
             .map(async |upstream| -> Result<Shared, Error> {
                 let pb = mp.insert_before(
                     &tp,
-                    ProgressBar::new(u64::MAX).with_prefix(format!(
+                    ProgressBar::new(u64::MAX).with_message(format!(
                         "{} {}",
                         "Downloading".blue(),
                         upstream.name().bold(),
--- a/boulder/src/upstream.rs
+++ b/boulder/src/upstream.rs
@@ -5,6 +5,7 @@
 use std::{io, path::Path, time::Duration};
 
 use crate::recipe::Recipe;
+use fs_err as fs;
 use futures_util::{StreamExt, TryStreamExt, stream};
 use moss::runtime;
 use stone_recipe::upstream;
@@ -34,7 +35,7 @@ pub enum Upstream {
 impl Upstream {
     /// Constructs an [Upstream] based on the information provided
     /// in the `upstream` section of a Stone recipe.
-    pub fn from_recipe_upstream(upstream: upstream::Upstream) -> Result<Self, Error> {
+    pub fn from_recipe_upstream(upstream: upstream::Upstream, original_index: usize) -> Result<Self, Error> {
         match upstream.props {
             upstream::Props::Plain { hash, rename, .. } => Ok(Self::Plain(Plain {
                 url: upstream.url,
@@ -44,6 +45,7 @@ impl Upstream {
             upstream::Props::Git { git_ref, .. } => Ok(Self::Git(Git {
                 url: upstream.url,
                 commit: git_ref,
+                original_index,
             })),
         }
     }
@@ -138,12 +140,18 @@ pub fn parse_recipe(recipe: &Recipe) -> Result<Vec<Upstream>, Error> {
         .upstreams
         .iter()
         .cloned()
-        .map(Upstream::from_recipe_upstream)
+        .enumerate()
+        .map(|(index, upstream)| Upstream::from_recipe_upstream(upstream, index))
         .collect()
 }
 
 /// Helper that stores and shares a list of [Upstream]s.
-pub fn sync(upstreams: &[Upstream], storage_dir: &Path, share_dir: &Path) -> Result<Vec<Shared>, Error> {
+pub fn sync(
+    recipe: &Recipe,
+    upstreams: &[Upstream],
+    storage_dir: &Path,
+    share_dir: &Path,
+) -> Result<Vec<Stored>, Error> {
     println!();
     println!("Sharing {} upstream(s) with the build container", upstreams.len());
 
@@ -157,9 +165,9 @@ pub fn sync(upstreams: &[Upstream], storage_dir: &Path, share_dir: &Path) -> Res
     );
     tp.tick();
 
-    let shared = runtime::block_on(
+    let stored = runtime::block_on(
         stream::iter(upstreams)
-            .map(async |upstream| -> Result<Shared, Error> {
+            .map(|upstream| async {
                 let pb = mp.insert_before(
                     &tp,
                     ProgressBar::new(u64::MAX).with_message(format!(
@@ -179,7 +187,7 @@ pub fn sync(upstreams: &[Upstream], storage_dir: &Path, share_dir: &Path) -> Res
                         .tick_chars("--=≡■≡=--"),
                 );
 
-                let shared = stored.share(share_dir).await?;
+                stored.share(share_dir).await?;
 
                 let cached_tag = stored
                     .was_cached()
@@ -191,16 +199,24 @@ pub fn sync(upstreams: &[Upstream], storage_dir: &Path, share_dir: &Path) -> Res
                 mp.suspend(|| println!("{} {}{cached_tag}", "Shared".green(), upstream.name().bold()));
                 tp.inc(1);
 
-                Ok(shared)
+                Ok(stored) as Result<_, Error>
             })
             .buffer_unordered(moss::environment::MAX_NETWORK_CONCURRENCY)
-            .try_collect(),
+            .try_collect::<Vec<_>>(),
     )?;
+
+    if let Some(updated_yaml) = update_git_upstream_refs(&recipe.source, &stored) {
+        fs::write(&recipe.path, updated_yaml)?;
+        println!(
+            "{} | Git references resolved to commit hashes and saved to stone.yaml. This ensures reproducible builds since tags and branches can move over time.",
+            "Warning".yellow()
+        );
+    }
 
     mp.clear()?;
     println!();
 
-    Ok(shared)
+    Ok(stored)
 }
 
 /// Helper that removes a list of [Upstream]s from the storage directory.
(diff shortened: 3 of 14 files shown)


crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+use std::path::{self, Path, PathBuf};
+use std::process::Stdio;
+use tokio::{
+    io,
+    process::{self},
+};
+use url::Url;
+
+pub mod error;
+pub use error::*;
```
tarkah: Please no `*` re-exports. We should only re-export items explicitly that are safe to do so w/ in the parent namespace. `Constraint` isn't one of those items... `error::Constraint` / error namespace is definitely needed.

tarkah: Please use `Std -> External -> Crate` import groupings / orderings. When re-exporting / using items from children modules, please group them in the `Crate` grouping and prefix their use with `self::`:

https://rust-lang.github.io/rustfmt/?version=v1.9.0&search=#group_imports

```rs
use std::path::Path;

use url::Url;

pub use self::error::Error
```

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+
+/// A Git repository.
+pub struct Repository {
+    path: PathBuf,
+}
+
+impl Repository {
+    /// Opens a local bare Git repository.
+    /// [Error::NotBare] is returned if it is not a bare repository.
```
tarkah: `Error::NotBare` isn't a doc link that'll resolve. Should it be `error::Constraint::NotBare`?

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        let output = run_git(&[
+            OsStr::new("-C"),
+            self.path.as_os_str(),
+            OsStr::new("cat-file"),
+            OsStr::new("-t"),
+            OsStr::new(commit),
+        ])
+        .await?;
+        Ok(output.stderr.is_empty())
+    }
```
tarkah, line 101: This technically will return true for any object type, not just commits.

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        .await
+        .map(|_| ())
+    }
+}
+
+/// The argument of callbacks when they are invoked
+/// for reporting a Git operation's progress.
+pub struct FetchProgress {
+    /// Isn't this self-explanatory?
+    pub percent: u8,
```
tarkah: Use the obvious comment xD `/// Completion percentage`

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        Ok(Self { path })
+    }
+
+    /// Clones a local or remote Git repository as bare into `path`.
+    /// A callback is fired repeatedly to track the cloning
+    /// process in real time.
+    pub async fn clone_bare_progress<F>(path: &Path, url: &Url, callback: F) -> Result<Self, Error>
+    where
+        F: Fn(FetchProgress),
+    {
```
tarkah: `clone_mirror` instead of `clone_bare`, since we use `--mirror`? Clone `--bare` means something different.

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+/// The argument of callbacks when they are invoked
+/// for reporting a Git operation's progress.
+pub struct FetchProgress {
+    /// Isn't this self-explanatory?
+    pub percent: u8,
+    /// Download speed in bytes per second.
+    pub speed: u64,
```
tarkah: Shouldn't we reuse the formatted / human readable `34.99 MiB/s`? It seems you're currently outputting raw bytes per second. The formatted progress bar will look better using this same human readable value.

boulder/src/upstream/git.rs
```diff
@@ -5,559 +5,213 @@
...
-    fn test_resolve_invalid_ref() {
-        let (temp_dir, _) = setup_test_repo();
-        let uri = Url::parse("https://example.com/test.git").unwrap();
+fn set_progress_bar_style(pb: &ProgressBar) -> impl Fn(gitwrap::FetchProgress) {
+    use tui::HumanBytes;
 
-        let err = resolve_git_ref(temp_dir.path(), "nonexistent", &uri).unwrap_err();
+    pb.set_length(100);
+    pb.set_style(
+        ProgressStyle::with_template("{prefix}\n|{bar:20.cyan/bue}| {percent}%, {msg}/s")
```
tarkah: We should right align & dim the human bytes / s like we do elsewhere... so format it w/ `/s` and then `{msg:>.dim} ` ?

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        let output = run_git(&[
+            OsStr::new("-C"),
+            path.as_os_str(),
+            OsStr::new("repo"),
+            OsStr::new("info"),
+            OsStr::new("layout.bare"),
+        ])
+        .await?;
+        if !output.stdout.starts_with(b"layout.bare=true") {
+            return Err(InnerError::Constraint(Constraint::NotBare))?;
```
tarkah, line 48: Is this guaranteed to be the first value always? vs using `contains`?

tarkah (commented):
I ran out of time, but here's my review so far. I still need to get eyes on how everything is hooked up in boulder.

crates/gitwrap/src/lib.rs
livingsilver94: I used `bare` because a repo may or may not be bare. Mirror is one way to clone a bare repo. Not a big deal though, I'll change it.

crates/gitwrap/src/lib.rs
livingsilver94: AFAIK nothing is guaranteed in the Git CLI. My experiments showed that the message is `layout.bare=true\n`. (<- well, or any other kind of combination to signal a new line).

I can switch to `contains` if we want to be paranoid about that, no problem.

crates/gitwrap/src/lib.rs
tarkah: Right, I think that makes sense for `open_bare`. But the cloning is specifically using `--mirror` and not `--bare` so its a bit misleading.

crates/gitwrap/src/lib.rs
livingsilver94: Good point! Thanks.

livingsilver94: `rustfmt`really, really needs to move faster to stabilize this stuff.

crates/gitwrap/src/lib.rs
livingsilver94: `contains()` takes one u8, not a subslice. To avoid complicating the logic (as there is no function in the stdlib to find subslices), I would assume `layout.bare=true\n` is always returned:

1. Because I'm running `git repo info layout.bare`, so the key is known a priori
2. Because `man git-repo` says 
    > "Retrieve metadata-related information about the current repository. Only the requested data will be returned based on their keys (see "INFO KEYS" section below)."

boulder/src/build.rs
```diff
@@ -128,14 +133,18 @@ impl Builder {
...
+        )?;
 
         timing.finish(timer);
 
-        Ok(())
+        Ok(shared)
     }
 
-    pub fn cleanup(&self) -> Result<(), Error> {
+    pub async fn cleanup(&self) -> Result<(), Error> {
```
tarkah: Why async? We're then calling non-async / blocking `fs` operations within this future. I don't think we need async fo any cleanup related functionality

tarkah: Ahh yeah, gitwrap's entire API is async. Can we have a `remove_sync` method on Workspace so we don't pollute all cleanup code w/ async and require spawning a new runtime just for this? 

This should be super doable with an inner `run_git` vs `run_git_sync` , the API is petty much identical between std & tokio process.

boulder/src/upstream.rs
```diff
@@ -183,47 +211,37 @@ pub fn sync(recipe: &Recipe, paths: &Paths, upstreams: &[Upstream]) -> Result<()
...
+    Ok(shared)
 }
 
-pub fn remove(paths: &Paths, upstreams: &[Upstream]) -> Result<(), Error> {
+/// Helper that removes a list of [Upstream]s from the storage directory.
+pub async fn remove(storage_dir: &Path, upstreams: &[Upstream]) -> Result<(), Error> {
     for upstream in upstreams {
-        upstream.remove(paths)?;
+        let stored = upstream.stored(storage_dir).await?;
+        stored.remove()?;
```
tarkah: We shouldn't have to recreate an entire `Stored` just to remove it. Can't we just have `Upstream::remove` -> `Git::remove`? We shouldn't need to run any git operations for this / do any validation we already did when storing it initially. Just remove the stored directory.

This + the other comment of having a `Worktree::remove_sync` function should allow us to eliminate all `async` boundaries / not require a new async runtime just to do simple FS cleanup operations.

boulder/src/cli/build.rs
```diff
@@ -105,7 +106,7 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
     );
     println!("boulder {}", tools_buildinfo::get_simple_version());
     println!("└─ building {pkg_name}-{build_release}\n");
-    builder.setup(&mut timing, timer, update)?;
+    let shared = builder.setup(&mut timing, timer, update)?;
```
tarkah: Nit:

```suggestion
    let shared_upstreams = builder.setup(&mut timing, timer, update)?;
```

boulder/src/cli/build.rs
```diff
@@ -148,9 +149,20 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
...
-        builder.cleanup().map_err(Error::Cleanup)?;
-    }
+    runtime::block_on(async {
+        // Unconditionally clean up shared sources...
+        for share in shared {
+            share
+                .remove()
+                .await
+                .map_err(|e| Error::Cleanup(build::Error::from(e)))?;
+        }
```
tarkah: Why are we unconditionally cleaning these up? What impact does this have on `chroot`?

I'd suggest we scope this under `cleanup` unless we fully understand the impacts to `chroot` and they're acceptable.

boulder/src/upstream/plain.rs
```diff
@@ -46,108 +58,167 @@ impl Plain {
...
+            name: self.name().to_owned(),
+            path,
+            was_cached: true,
+        })
+    }
+
+    /// Returns a relative PathBuf where this source archive
+    /// should be stored within the storage directory.
+    pub fn stored_path(&self, storage_dir: &Path) -> PathBuf {
+        [storage_dir, &self.file_path()].iter().collect()
```
tarkah:
```suggestion
    pub fn stored_path(&self, storage_dir: &Path) -> PathBuf {
        storage_dir.join(self.file_path())
```

boulder/src/upstream/plain.rs
```diff
@@ -46,108 +58,167 @@ impl Plain {
...
+        let path = self.stored_path(storage_dir);
 
-        fs::remove_file(&path)?;
-
-        if let Some(parent) = path.parent() {
-            util::remove_empty_dirs(parent, &paths.upstreams().host)?;
+        let mut file = fs_err::File::open(&path)?;
+        let mut hasher = Sha256::new();
+        io::copy(&mut file, &mut hasher)?;
+        let hash = hex::encode(hasher.finalize());
```
tarkah: `util::sha256_hash`

boulder/src/upstream/plain.rs
```diff
@@ -46,108 +58,167 @@ impl Plain {
...
+        if let Err(err) = &result
+            && err.kind() == io::ErrorKind::CrossesDevices
+        {
+            // Source and destination paths
+            // reside on different filesystems.
+            // Copy it instead.
+            fs::copy(&self.path, &target).map(|_| ())
+        } else {
+            result
+        }?;
```
tarkah: `util::hardlink_or_copy`

boulder/src/upstream/plain.rs
```diff
@@ -46,108 +58,167 @@ impl Plain {
...
+        fs::remove_file(&self.path)?;
+
+        let parents = self.path.parent().unwrap_or(Path::new("")).iter();
+        for parent in parents.rev() {
+            match fs::remove_dir(parent) {
+                Ok(_) => {}
+                Err(e) if e.kind() == io::ErrorKind::DirectoryNotEmpty => break,
+                Err(e) => return Err(Error::from(e)),
+            }
+        }
```
tarkah: `util::remove_empty_dirs`

I also don't think the current implementation will work since you're iterating over the segments of a path, not the full path w/ each trailing segment removed as you iterate.

boulder/src/upstream.rs
```diff
@@ -4,130 +4,166 @@
...
+    /// or as a directory depending on the kind of source.
+    ///
+    /// Information will be inferred according to the kind of source URI.
+    /// See variants of [Upstream] for more details.
     pub async fn _fetch_new(uri: SourceUri, dest: &Path) -> Result<Self, Error> {
         Ok(match uri.kind {
             upstream::Kind::Archive => Self::Plain(Plain::_fetch_new(uri.url, dest).await?),
             upstream::Kind::Git => Self::Git(Git::_fetch_new(&uri.url, dest).await?),
         })
     }
```
tarkah: Are these `_fetch_new` methods needed anymore?

boulder/src/upstream.rs
```diff
@@ -4,130 +4,166 @@
...
-    async fn store(&self, paths: &Paths, pb: &ProgressBar) -> Result<Stored, Error> {
+    /// Stores the upstream into the storage directory.
+    /// The final path contained in the storage directory, and the write logic,
+    /// depend on the upstream variant. The final path where the upstream is stored
+    /// is unique inside the storage directory.
+    async fn store(&self, storage_dir: &Path, pb: &ProgressBar) -> Result<Stored, Error> {
         Ok(match self {
-            Upstream::Plain(plain) => Stored::Plain(plain.store(paths, pb).await?),
-            Upstream::Git(git) => Stored::Git(git.store(paths, pb).await?),
+            Upstream::Plain(plain) => Stored::Plain(plain.store(storage_dir, pb).await?),
```
tarkah, line 69: Aren't we missing `.join("fetched")` here?

Also, we're leaking the implementation detail of this constructed "root" to this module. `.join("git") / .join("fetched")` from the storage root should happen within each child module, not at the parent level here.

boulder/src/upstream/git.rs
```diff
@@ -5,554 +5,213 @@
...
+            },
+            has_ref,
+        ))
+    }
+
+    /// Returns a relative PathBuf where this Git repository
+    /// should be stored within the recipe storage.
+    pub fn _stored_path(&self, storage_dir: &Path) -> PathBuf {
+        [storage_dir, &self.directory_name()].iter().collect()
+    }
```
tarkah, line 107: Dead code

boulder/src/upstream/git.rs
```diff
@@ -5,554 +5,213 @@
...
+                }
+                repo
+            }
+            Err(e) => {
+                if e.run_failed() {
+                    cached = false;
+                    clone(&self.url, &dir, pb).await?
+                } else {
+                    return Err(Error::from(e));
+                }
```
tarkah: Wont this bail if `Constraint::NotBare` is returned? Why is this unrecoverable? 

I'd suggest we just always remove `dir` and then `clone` if its not a valid bare git repo already.

boulder/src/upstream/git.rs
```diff
@@ -5,554 +5,213 @@
...
-        if let Stored::Git(git) = stored
-            && git.resolved_hash != git.original_ref
+    /// Removes the stored Git repository.
+    /// Should the stored repository no longer exist,
+    /// this function returns successfully (it is idempotent).
+    pub fn remove(&self) -> Result<(), Error> {
+        let result = fs::remove_dir_all(self.repo.path());
+        if let Err(err) = result
+            && err.kind() != io::ErrorKind::NotFound
         {
```
tarkah: `util::remove_dir_all`

boulder/src/cli/build.rs
ermo: Just writing here to say that the ability to `chroot` in is used frequently by e.g. Reilly.

Killing chroot ability by default is not good UX for us. Is there a way we can avoid that, perhaps in the form of scoping unconditional cleanup to the --cleanup flag alone?

Fair dos if we need to have a separate boulder clean command (similar to e.g. `solbuild dc`).

boulder/src/upstream.rs
livingsilver94: I left them here because I may use them when I'll add the ability to `boulder recipe new` a Git upstream. I'm not sure I'll do that though, that was just my mental scaffolding. I can remove the (currently) dead code, I'll keep such code saved locally should I really need it.

ermo: Maybe leave them in, but commented out as a `// TODO: boulder recipe new (some git upstream)`?

boulder/src/upstream/plain.rs
livingsilver94: The doc says it's more efficient to collect (without giving numbers) so I just assumed it's the best practice.

boulder/src/upstream.rs
livingsilver94: My idea was to honor invariants: you can't delete anything if you can't prove you stored it first.

boulder/src/upstream/plain.rs
livingsilver94: > you're iterating over the segments of a path

Dang, I missed that. Definitely broken, thanks.

boulder/src/upstream.rs
livingsilver94: Oh no no, Joey will promptly nuke the TODO comments. He's *eager*! /j

boulder/src/upstream/git.rs
livingsilver94: > Wont this bail if Constraint::NotBare is returned? Why is this unrecoverable?

Yes. I thought it's better to bail if something/someone tampered with the storage. What's the assumption here? We own the directory so we're free to do anything we want inside it?

boulder/src/upstream.rs
livingsilver94: > `.join("git") / .join("fetched")` from the storage root should happen within each child module

Fine, I changed that.

Do note though that with this PR the `fetched` directory is no more, as I see no benefit in it. Here's the tree:

```text
~/.cache/boulder
└── upstreams
    └── git
        └── invent.kde.org..qt.qt.qtdeclarative.git
```

Yes I know this is a breaking change, but why should we retain `fetched`?

boulder/src/cli/build.rs
livingsilver94: Yeah, I did not consider the `choot` thingy, I'm moving the new code inside the `if cleanup` path.

But I also think the cleanup is too coarse since for example it deletes upstreams too, whereas I may only want to clean old build environments. Am I getting this right?

ermo: We have the "full" cleanup because we use it with the build agents, so they never run out of space. This means they are completely hands-off once they are up and running (which is a huge win IME).

We have discussed more fine-grained cache deletion as well.

This may be out of scope of the current PR, but I can imagine that `--cleanup` might take arguments (and have a default argument of "all"), so that you can clean different parts of the build cache.

boulder/src/upstream/plain.rs
tarkah: If anything it's a premature optimization. Join is much easier to read.

boulder/src/upstream.rs
tarkah: Why introduce a breaking change and nuke everyone's cached upstreams? It doesn't hurt having the extra directory, so I think we need a stronger argument to remove it than keep it.

livingsilver94: I've updated the cover letter to better highligh the changes in paths. Directories of git repositories are now named after their name + the url, so that alone is a breaking change. IIRC this was a desired feature to avoid collisions. Removing the `fetched` directory is just a nit.

boulder/src/cli/build.rs
livingsilver94: > This may be out of scope of the current PR

It surely is! But I think it's worth having. I may file another PR after this one is merged. This one is huge enough already.

boulder/src/upstream/git.rs
```diff
@@ -5,554 +5,200 @@
...
+    /// directory, that would store the Git repository.
+    /// If the directory does not exist, this function returns
+    /// successfully (it is idempotent).
+    ///
+    /// Careful: this function does not validate the content
+    /// of the directory! Resources will be deleted even if they
+    /// do not belong to a Git repository.
+    pub fn remove(&self, storage_dir: &Path) -> Result<(), Error> {
+        let dir = storage_dir.join(self.directory_name());
+        util::remove_dir_all(&dir).map_err(Error::from)
```
tarkah: Don't forget to use proper `stored_path` helper. After this change, I've confirmed `directory_name` is only called from `stored_path`

```suggestion
        let dir = self.stored_path(storage_dir);
        util::remove_dir_all(&dir).map_err(Error::from)
```

crates/gitwrap/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+            OsStr::new("cat-file"),
+            OsStr::new("-t"),
+            OsStr::new(commit),
+        ])
+        .await?;
+        if !output.stdout.starts_with(b"commit") {
+            return Err(InnerError::Constraint(Constraint::NotPeeled {
+                commit: commit.to_owned(),
+            }))?;
+        }
```
tarkah: May as well use our helper here

```suggestion
        if !self.has_commit(commit).await? { {
            return Err(InnerError::Constraint(Constraint::NotPeeled {
                commit: commit.to_owned(),
            }))?;
        }
```

boulder/src/upstream.rs
tarkah: Please add back `fetched` prefix to `plain` upstream paths. We scope git upstreams under `git`. There's no benefit to breaking peoples caching and polluting the root `upstreams` cache folder with a bunch of hash prefixes, burying the `git` folder in that noise.

tarkah (changes requested):

tarkah (approved):
LGTM

ermo:
@livingsilver94 I'm not seeing any test plan for this that I can attempt to reproduce locally? 

Hence, I would like to ask which kind of tests you have run so far to satisfy yourself that this is not likely to have regressions when used like a normal packager would use it...?

ermo (approved):
Merging per @tarkah's review.

@livingsilver94 Thank you for sharing the Test Plan. That was very helpful.

@ReillyBrogan Just pinging you to ensure that you are brought into the loop on this one, in case you manage to hit a corner case after this has been merged into prod. In particular, note that support for the `staging` property field has now been removed, as it is allegedly no longer necessary when using the sharing approach this PR adds.

PR #755, opened 2026-04-02
moss: Add support for simulating privileged operations

joebonrichie:
Currently, it is effectively just an early return, but, it can be helpful to ensure dependency trees.

This is preliminary groundwork for the moss packagekit backend as packagekit always simulates operations first and we want the end high-level API to support this.

### DIFF
--- a/moss/src/cli/install.rs
+++ b/moss/src/cli/install.rs
@@ -26,6 +26,10 @@ pub struct Command {
     /// Packages to install
     packages: Vec<String>,
 
+    /// Simulate the operation (dry-run)
+    #[arg(long)]
+    dry_run: bool,
+
     /// Blit this sync to the provided directory instead of the root
     ///
     /// This operation won't be captured as a new state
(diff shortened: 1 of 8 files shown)


tarkah (approved):
LGTM. What do we need to eventually output in simulation mode for proper packagekit integration?

moss/src/cli/install.rs
```diff
@@ -4,49 +4,57 @@
...
+    about = "Install packages",
+    long_about = "Install packages by name"
+)]
+pub struct Command {
+    /// Packages to install
+    packages: Vec<String>,
+
+    /// Simulate the operation (dry-run)
+    #[arg(short, long)]
+    simulate: bool,
```
tarkah: Should we call this --dry-run instead? I feel that's more common than --simulate

livingsilver94: 100% agree. If there are "standard" flags out there, they certainly are `-v` and `--dry-run`.

joebonrichie: yep, agreed

tarkah (approved):

PR #759, opened 2026-04-07
moss: UX & perf. improvings for removing stale states

joebonrichie:
Add tracing and progress bar output when removing states

Use a parallelized recursive function that can remove all the stale states in one call

On an NVME drive on a modern system for the same pkg lists (hot) Before: 0m3.364s Now: 0m2.284s

### DIFF
--- a/moss/src/client/prune.rs
+++ b/moss/src/client/prune.rs
@@ -9,6 +9,9 @@
 //! and assets on disk by way of refcounting.
 
 use std::collections::{BTreeMap, BTreeSet};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::{Duration, Instant};
 use std::{
     io,
     path::{Path, PathBuf},
@@ -18,6 +21,8 @@ use fs_err as fs;
 use itertools::Itertools;
 use thiserror::Error;
 
+use tracing::info;
+use tui::{ProgressBar, ProgressStyle};
 use tui::{
     dialoguer::{Confirm, theme::ColorfulTheme},
     pretty::autoprint_columns,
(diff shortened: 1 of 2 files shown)


ermo (commented):
Is there a way to coalesce this so we only create one thread pool and then let that delete *everything* (all the relevant states) in parallel?

So basically "give the function a vec of (refs to) dirs that need to be recursively deleted"?

I hope what I am attempting to convey makes sense.

So instead of `[some list of state dirs].map(|d| remove_parallel(d))`, do `remove_parallel([some list of state dirs])`.

joebonrichie:
Yep, definitely could. It'll probably be quicker than re-creating the threadpool as you surmise. I was actually thinking of adding some progress bar reporting around this for loop though, currently something like `state prune` takes a long time with zero progress indicated to the user. So it'll be a tradeoff between this unless we reuse the threadpool instead of recreating it.

ermo:
My suggestion would be to track the top-level dirs and report a callback when each of those have been removed (so tracking arg-completion), and then otherwise have the "heartbeat" timer pulse?

ermo (commented):
At a glance, this looks very nice.

moss/src/client/prune.rs
```diff
@@ -168,14 +192,54 @@ pub(super) fn prune_states(client: &Client, strategy: Strategy<'_>, yes: bool) -
...
+    );
+
+    let progressbar = ProgressBar::new(archive_paths.len() as u64)
+        .with_style(ProgressStyle::with_template("\n|{bar:20.cyan/blue}| {pos}/{len}").unwrap());
+
+    util::par_remove_dirs_all(archive_paths.iter().map(|p| p.as_path()).collect(), |path, res| {
+        let mut counter = 0;
+        match res {
+            Ok(_) => {
+                counter += 1;
```
tarkah: This `counter` will always be `1`, its local to this closure and not shared across all parallelized callbacks

tarkah (approved):

PR #715, opened 2026-02-25
.github: Check REUSE compliance in parallel to code conformity

livingsilver94:
For transparency, I used this script to replace the current copyright header with the format we agreed upon (that is `SPDX-FileCopyrightText: <file_creation_year> AerynOS Developers`)

```shell
#!/bin/sh

files=$(find . -type f -not -path './.git/*')
for f in $files; do
    echo $f
    year=$(date -d @$(git log --reverse --follow --format='%at' "$f" | head -n1) +%Y)
    sed "s|\(.*SPDX-FileCopyrightText: \).*|\1$year AerynOS Developers|" -i "$f"
done
```

This PR solves #714.

PR #761, opened 2026-04-08
Achieve REUSE compliance

livingsilver94:
I'm replacing PR #715 with this one, as the former was created when I had no write permission on this project. (If only Github allowed to change the source branch of a PR...)

This PR solves #714.


livingsilver94:
@ermo this PR should be ready. The changes introduced here are a replica of what done in https://github.com/AerynOS/recipes/pull/1524.

PR #764, opened 2026-04-09
boulder/cli: Add cache {clean,size} subcommand options

joebonrichie:
Fairly self explanatory

### DIFF
--- a/boulder/src/cli/cache.rs
+++ b/boulder/src/cli/cache.rs
@@ -1,12 +1,15 @@
 // SPDX-FileCopyrightText: 2026 AerynOS Developers
 // SPDX-License-Identifier: MPL-2.0
 
-use std::{io, path::PathBuf};
+use std::{
+    io,
+    path::{Path, PathBuf},
+};
 
 use clap::{Args, Parser};
 use container::Container;
 use humansize::BINARY;
-use moss::{Installation, util};
+use moss::util;
 use rayon::iter::{ParallelBridge, ParallelIterator};
 use thiserror::Error;
 use walkdir::WalkDir;
@@ -55,13 +58,14 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
 }
 
 fn clean(env: Env, boulder_cache: bool, moss_cache: bool) -> Result<(), Error> {
-    let installation = Installation::open(&env.moss_dir, None)?;
-
     for (name, path) in selected_caches(&env, boulder_cache, moss_cache) {
+        let tmpdir = tempfile::tempdir()?;
+
         println!("Deleting {name} directory: {}", path.display());
-        Container::new(&installation.root)
-            .bind_rw(&path, &path)
-            .run(|| util::par_remove_dir_all(&path))?;
+
+        Container::new(tmpdir.path())
+            .bind_rw(&path, Path::new("/remove"))
+            .run(|| util::par_remove_dir_all(Path::new("/remove")))?;
     }
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


ermo:
This looks pretty neat.

ermo:
I think `boulder cache clean` would work nicely. That would make the aliases `cc` for clean and `cs` for size. The context is that we're not removing (or deleting) the cache, we're just cleaning it up.

boulder/src/cli/cache.rs
```diff
@@ -0,0 +1,100 @@
...
+        global = true
+    )]
+    pub boulder_cache: bool,
+    #[arg(short, long, help = "Delete the moss cache", default_value_t = false, global = true)]
+    pub moss_cache: bool,
+}
+
+#[derive(Debug, clap::Subcommand)]
+pub enum Subcommand {
+    #[command(about = "Delete the cache(s) for the current environment")]
```
ermo:
```suggestion
    #[command(about = "Clean out the cache(s) for the current environment")]
```

ermo (commented):
Tiny nit.

boulder/src/cli/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::{io, path::PathBuf};
+
+use clap::{Args, Parser};
+use humansize::BINARY;
+use rayon::iter::{ParallelBridge, ParallelIterator};
+use thiserror::Error;
+use walkdir::WalkDir;
+
+use container::Container;
+use moss::{Installation, util};
```
tarkah: Nit: no space on external crates & sort

boulder/src/cli/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+
+fn clean(env: Env, boulder_cache: bool, moss_cache: bool) -> Result<(), Error> {
+    let installation = Installation::open(&env.moss_dir, None)?;
+
+    for (name, path) in selected_caches(&env, boulder_cache, moss_cache) {
+        print!("Deleting {name} directory: {}", path.display());
+        Container::new(&installation.root)
+            .bind_rw(&path, &path)
+            .work_dir(&path)
+            .run(|| util::par_remove_dir_all(&path))?;
```
tarkah: Whats the relationship between `Installation.root` and the boulder cache dir? They might not overlap at all?

I think this is just coincidentally working if `installation.root` is `/`. Otherwise the container root wont have `path` under it for either since its relative to `/` and `installation.root` of something else wont resolve to that?

tarkah: Also do we need to be in a container to cleanup moss files? I guess transaction triggers on archived roots might cause an issue, but boulder doesn't utilize moss for that.

joebonrichie: So we're using env.moss_dir as the root here, which is typically ~/.cache/moss. This is the same directory that boulder profile update uses for example, so it should always be available assuming a profile is initialized. 

We do need to be in a container specifically for some files in the moss isolation directory as they can be unremovable without root (a tmpfile in the colord package is a known example) going inside the namespace is a workaround for that so boulder can be ran without root. We could delete without a container on a best-effort run and simply accept that some files won't be able to be deleted without root.

tarkah: Right but when env.moss_dir is used as the container root, `env.moss_dir/.moss` won't be a reachable path anymore. It'd now be `/.moss` within the container.

tarkah: Oh jfc I missed the bind. We can remove the bind if you do what I mention, but I guess the bind fixes this.

tarkah (approved):

tarkah (approved):

PR #767, opened 2026-04-16
moss/cli: Add preliminary root check before managing boot items

joebonrichie:
Currently, we allow any permissions issues to fail at execution time, however, boot status will report:

`XBOOTLDR       : None`

Even when an XBOOTLDR partition is present unless ran with root.

Therefore do a prelim root check on this code path


joebonrichie:
Close for now, will work on better .is_root() integration repo-wide

PR #768, opened 2026-04-19
crates: gitwrap: Achieve REUSE compliance

livingsilver94:
This PR just makes REUSE checks pass.

While we were developing gitwrap, we in parallel introduced checks for REUSE compliance. I did not feel like adding a new commit to the PR of gitwrap (#761) because it was pretty big and controversial for some aspects. I wanted it to be stable for the sake of the review process.

Now that we merged both the REUSE checks and gitwrap, it's time to make it pass REUSE checks.


ermo (approved):
LGTM, thanks!

PR #769, opened 2026-04-20
moss/cli: Nuke list sync support

joebonrichie:
Now handled by `moss sync --dry-run`

CC @AerynOS/web-staff for documentation

### DIFF
--- a/moss/src/cli/list.rs
+++ b/moss/src/cli/list.rs
@@ -29,39 +29,18 @@ pub fn command() -> Command {
                 .about("List all available packages")
                 .visible_alias("la"),
         )
-        .subcommand(
-            Command::new("sync")
-                .about("List packages with sync changes")
-                .visible_aliases(["ls", "lu"])
-                .arg(arg!(--"upgrade-only" "Only sync packages that have a version upgrade")),
-        )
-}
-
-enum Sync {
-    All,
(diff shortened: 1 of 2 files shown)


CookieSource:
Thanks for the ping I will open an issue on our side to process it!

tarkah (approved):
Nice! This eliminates some unnecessary complexity. I'm aligned w/ this.

ermo:
@joebonrichie Having thought about this, and having subsequently had a discussion with tarkah about it, I think I am going to have to rescind my blessing for this, for which I apologise.

The reason is that, in the glorious CLI future we are moving toward, we are going to have the `list` verb be `moss pkg/package list` and `moss repo list`. If we also then support JSON for those two verbs, that will *only* output packages that can be updated, similar to how you enabled the `ent` output to be shown as pure JSON with no embellishments.

In comparison, --dry-run is conceptually different, in that it should basically "mock" the entire action except from actually committing the changes to disk. In this context, the `moss sync --dry-run` output is a side effect, not the main purpose, so it does not really make sense to have the `--dry-run` output supersede the `list sync` output.

Does that make sense to you?

ermo:
Closing per reasoning above. Sorry again for the dithering on this one, @joebonrichie

PR #773, opened 2026-04-22
Task/recipe update polish

tarkah:
Makes diff output colorized (while still `diff -u` compatible) and hides some debuggy output behind `verbose` flag

image


ermo (approved):
LGTM, thanks!

PR #777, opened 2026-05-04
moss: Add testing flag to dev-deps

tarkah:
Adds `testing` as a `dev-dependencies` so one doesn't need `--features testing` for `cargo test` to work. It works by default now.


ermo (approved):
\o/

PR #787, opened 2026-05-18
Migrate from Diesel to rusqlite

livingsilver94:
This PR moves moss/db code from using the Diesel ORM to pure SQLite using the rusqlite wrapper. It does so by first introducing a battery of unit tests, making them pass with Diesel, and then using them as the guideline to perform the conversion.

rusqlite is a thin wrapper around SQLite's API so it lacks a schema migration functionality. We currently don't need a complex one, so I wrote our own logic, tested by a small battery of tests as well.

I tried not to change the API; the changes that I introduced are minimal al all contained within the moss codebase. I think some portions of the code could be improved, so I added TODO or FIXME comments as memos. I encourage you to check those comments if you don't want them committed.

~~I've always been told to reduce the number of queries to get data. This is especially true for DMBS over the network and not-so-true for SQLite, still I think it's a good habit. At any rate, for one-to-many relationships I decided to build the list of objects returned by single queries as a JSON array, so that it's easy to parse on moss' side. So yes, this PR also introduces a dependency on serde_json.~~
This turned out to be detrimental with SQLite. It's better to spawn multiple (and simpler) queries.


livingsilver94:
Unit tests pass but I haven't run `moss` neither with an existing DB nor a new one, hence the draft.

livingsilver94:
I think we could use a minifier for SQL code to reduce the final size of the `moss` executable.

moss/src/db/layout/mod.rs
```diff
@@ -40,165 +41,113 @@ impl Database {
...
-                );
-            }
-
-            Ok(output)
+            let ids: Rc<Vec<Value>> = Rc::new(packages.into_iter().map(|id| Value::from(id.to_string())).collect());
+            let mut stmt = conn.prepare("SELECT * FROM layout WHERE package_id IN rarray(?)")?;
+            Ok(stmt
+                .query_map([ids], |row| {
+                    let id = row.get::<_, String>("package_id")?.into();
+                    let layout = TryInto::<Layout>::try_into(row)?.into_inner();
```
tarkah, line 54: Nifty, loading the array to a vtable then using an efficient IN btree set filter.

Are we still limited to  `MAX_VARIABLE_NUMBER`? Or does this somehow count as a single bind? It appears it is passing the RC pointer to sqlite so it appears this is safe 👍

moss/src/db/layout/mod.rs
```diff
@@ -40,165 +41,113 @@ impl Database {
...
+                        l.1.uid,
+                        l.1.gid,
+                        l.1.mode,
+                        l.1.tag,
+                        file.0,
+                        file.1,
+                        file.2,
+                    );
+                    stmt.execute(values)?;
+                }
```
tarkah: We're no longer batching `MAX_VARIABLE_NUMBER / 8` chunks which is a performance regression. We should be able to format the binds for each chunk / add them.

moss/src/db/layout/mod.rs
```diff
@@ -40,165 +41,113 @@ impl Database {
...
-                        entry_value1,
-                        entry_value2,
-                    }
-                })
-                .collect::<Vec<_>>();
-
-            ids.sort();
-            ids.dedup();
-            batch_remove_impl(&ids, tx)?;
-
```
tarkah, line 132: This introduces a huge bug by not upholding that `package_id` should effectively be unique. 

We either need to remove -> add like previously, or add an `ON CONFLICT DO UPDATE SET` "upsert clause" https://sqlite.org/lang_upsert.html to override existing package_id with the incoming values. In theory we could also use `ON CONFLICT DO NOTHING` since it _should_ be the same data for a `package_id`, but lets not introduce any functionality change here.

moss/src/db/layout/mod.rs
```diff
@@ -40,165 +41,113 @@ impl Database {
...
-            Ok(())
+    pub fn batch_remove<'a>(&mut self, packages: impl IntoIterator<Item = &'a package::Id>) -> Result<(), Error> {
+        self.conn.exec_mut(|conn| {
+            let tx = conn.transaction()?;
+            {
+                let mut stmt = tx.prepare("DELETE FROM layout WHERE package_id = ?")?;
+                for p in packages {
+                    stmt.execute([p.as_str()])?;
+                }
```
tarkah: Same here, we're no longer actually batching the deletes. Can't we use `rarray` w/ `WHERE package_id IN (?)` like we do above? Then we don't even need to chunk within `MAX_VARIABLE_NUMBER`

moss/src/db/layout/mod.rs
```diff
@@ -212,73 +161,153 @@ fn encode_entry(entry: &StonePayloadLayoutFile) -> (&'static str, Option<Cow<'_,
...
+        Database::new(":memory:").map(|_| ())
+    }
 
-    use super::*;
+    #[test]
+    fn db_queries_package_id() -> Result<(), Error> {
+        let non_unique_id = package::Id::from(NUM_ENTRIES.to_string());
+        let entries = SHARED_DB.query(iter::once(&non_unique_id))?;
+        let expected_entries = all_entries().filter(|(id, _)| id == &non_unique_id);
+        itertools::assert_equal(entries.into_iter(), expected_entries);
```
tarkah: Theres some clippy warnings on this line, you can remove `into_iter` since it takes an IntoIterator

moss/src/db/layout/mod.rs
```diff
@@ -212,73 +161,153 @@ fn encode_entry(entry: &StonePayloadLayoutFile) -> (&'static str, Option<Cow<'_,
...
+    static SHARED_DB: LazyLock<Database> = LazyLock::new(|| {
+        // FIXME: This allocates a lot. Maybe we can make batched functions
+        // in mod.rs accept Borrow, so that it works with both owned and referenced values?
+        let entries = all_entries().collect::<Vec<_>>();
+        let entries_ref = entries.iter().map(|(id, rec)| (id, rec)).collect::<Vec<_>>();
+
+        let mut db = Database::new(":memory:").unwrap();
+        db.batch_add(entries_ref).unwrap();
+        db
+    });
```
tarkah: Why do we need a process shared DB? Its even better if each test has its own isolated DB. This will also "break" if we use something like `nextest` which compiles / runs each test as an isolated process. As in its no longer really shared and if tests rely on that behavior, it'll break that guarantee.

moss/src/db/layout/mod.rs
```diff
@@ -212,73 +161,153 @@ fn encode_entry(entry: &StonePayloadLayoutFile) -> (&'static str, Option<Cow<'_,
...
+    }
 
-        assert_eq!(count, all.len());
+    const NUM_ENTRIES: u32 = 512;
+
+    static SHARED_DB: LazyLock<Database> = LazyLock::new(|| {
+        // FIXME: This allocates a lot. Maybe we can make batched functions
+        // in mod.rs accept Borrow, so that it works with both owned and referenced values?
+        let entries = all_entries().collect::<Vec<_>>();
+        let entries_ref = entries.iter().map(|(id, rec)| (id, rec)).collect::<Vec<_>>();
```
tarkah: Whats "a lot"? These are tests and 512 DB entries shouldn't be that much data?

moss/src/db/meta/mod.rs
```diff
@@ -1,562 +1,540 @@
...
     }
 
     pub fn wipe(&self) -> Result<(), Error> {
-        self.conn.exclusive_tx(|tx| {
-            // Cascading wipes other tables
-            diesel::delete(model::meta::table).execute(tx)?;
-            Ok(())
-        })
+        self.conn
+            .exec(|conn| Ok(conn.execute("DELETE FROM meta", []).map(|_| ())?))
```
tarkah, line 54: I believe we need to enable foreign key pragma, otherwise the foreign key constraints in the schema during migration might fail & we definitely wont get cascading, which this relies on to cleanup the relation table entries.

Diesel must have done this automagically for us, but I believe we will need to enable that pragma prior to executing migrations for both `state` and `meta`. `layout` doesn't have any foreign keys defined.

moss/src/db/meta/mod.rs
```diff
@@ -1,562 +1,540 @@
...
+        json_group_array(DISTINCT md.dependency)
+            FILTER (WHERE md.dependency IS NOT NULL) AS dependencies,
+        json_group_array(DISTINCT mp.provider)       AS providers,
+        json_group_array(DISTINCT mc.conflict)
+            FILTER (WHERE mc.conflict IS NOT NULL)   AS conflicts
+    FROM meta m
+    JOIN meta_licenses ml ON m.package = ml.package
+    LEFT JOIN meta_dependencies md ON m.package = md.package
+    JOIN meta_providers mp ON m.package = mp.package
+    LEFT JOIN meta_conflicts mc ON m.package = mc.package
```
tarkah: I'm not so sure about this query. This feels very suboptimal from a query execution standpoint. And the multiple unrelated 1:N joins makes the result set grow multiplicatively.

 If we really want to utilize a single query w/ json_group_array, we probably want something more like this (query not checked for accuracy). Where `Database::query` doesn't try to share the same underlying query & instead uses something more optimized for an initial CTE that filters down the package set, which the subsequent CTEs and final query can join on via `package`.

```sql
WITH licenses AS (
    SELECT
      package,
      json_group_array(DISTINCT license) AS licenses
    FROM meta_licenses
    WHERE package = :package
    GROUP BY package
), dependencies AS (
    SELECT
      package,
      json_group_array(DISTINCT dependency) AS dependencies
    FROM meta_dependencies
    WHERE package = :package
    GROUP BY package
), providers AS (
    SELECT
      package,
      json_group_array(DISTINCT provider) AS providers
    FROM meta_providers
    WHERE package = :package
    GROUP BY package
), conflicts AS (
    SELECT
      package,
      json_group_array(DISTINCT conflicts) AS conflicts
    FROM meta_conflicts
    WHERE package = :package
    GROUP BY package
)
SELECT
    m.*,
    ml.licenses,
    md.dependencies,
    mp.providers,
    mc.conflicts
FROM meta m
  LEFT JOIN licenses ml USING (package)
  LEFT JOIN dependencies md USING (package)
  LEFT JOIN providers mp USING (package)
  LEFT JOIN conflicts mc USING (package)
WHERE m.package = :package
```

moss/src/db/meta/mod.rs
```diff
@@ -1,562 +1,540 @@
...
 
     pub fn file_hashes(&self) -> Result<BTreeSet<String>, Error> {
         self.conn.exec(|conn| {
-            Ok(model::meta::table
-                .select(model::meta::hash.assume_not_null())
-                .filter(model::meta::hash.is_not_null())
-                .distinct()
-                .load_iter::<String, _>(conn)?
-                .collect::<Result<_, _>>()?)
+            let mut stmt = conn.prepare("SELECT hash FROM meta WHERE hash IS NOT NULL")?;
```
tarkah: Nit: `DISTINCT` can be added to the query to encode the guarantee to the DB layer, even though we collect it into a BTreeSet anyways.

moss/src/db/meta/mod.rs
```diff
@@ -1,562 +1,540 @@
...
-                        (
-                            model::meta_conflicts::package.eq(package.as_str()),
-                            model::meta_conflicts::conflict.eq(conflict.to_string()),
-                        )
-                    })
-                })
-                .collect::<Vec<_>>();
-
-            batch_remove_impl(&ids, tx)?;
-
```
tarkah, line 324: Same as `layout`, we need to ensure added packages are upserted. `INSERT OR REPLACE` seems good for this, but we're not handling the relation tables at all, and will simply append new values / duplicate them.

Also we're no longer doing any batching here like before.

moss/src/db/meta/mod.rs
```diff
@@ -1,562 +1,540 @@
...
-            Ok(())
+    pub fn batch_remove<'a>(&mut self, packages: impl IntoIterator<Item = &'a package::Id>) -> Result<(), Error> {
+        self.conn.exec_mut(|conn| {
+            let tx = conn.transaction()?;
+            {
+                let mut stmt = tx.prepare("DELETE FROM meta WHERE package = ?")?;
+                for package in packages {
+                    stmt.execute((package.to_string(),))?;
+                }
+            }
```
tarkah: We're no longer batching removes. We can use the `rarray` method here.

moss/src/db/state/mod.rs
```diff
@@ -2,272 +2,347 @@
...
-                // Cascading wipes other tables
-                diesel::delete(model::state::table.filter(model::state::id.eq_any(chunk))).execute(tx)?;
+    pub fn batch_remove<'a>(&mut self, ids: impl IntoIterator<Item = &'a Id>) -> Result<(), Error> {
+        self.conn.exec_mut(|conn| {
+            let tx = conn.transaction()?;
+            {
+                let mut stmt = tx.prepare("DELETE FROM state WHERE id = ?")?;
+                for id in ids {
+                    stmt.execute((id.to_string(),))?;
+                }
```
tarkah: Like comments elsewhere, we should still batch this. We can use the `rarray` method here.

moss/src/db/state/mod.rs
```diff
@@ -2,272 +2,347 @@
...
-                        package_id: selection.package.as_str(),
-                        explicit: selection.explicit,
-                        reason: selection.reason.as_deref(),
-                    })
-                    .collect::<Vec<_>>();
-
-                for chunk in selections.chunks(MAX_VARIABLE_NUMBER / 4) {
-                    diesel::insert_into(model::state_selections::table)
-                        .values(chunk)
-                        .execute(tx)?;
```
tarkah, line 148: Ideally selections are batched as well. A state can have thousands of packages and we're now going from `O(CEIL(N / (32766 / 4)))` to `O(N)` in inserts

moss/src/db/state/mod.rs
```diff
@@ -2,272 +2,347 @@
...
+            let mut stmt = conn.prepare(indoc! {"
+                SELECT
+                    s.id, s.summary, s.description, s.created, s.type,
+                    json_group_array(
+                        json_object(
+                            'package_id', ss.package_id,
+                            'explicit',   ss.explicit,
+                            'reason',     ss.reason
+                        )
+                    ) FILTER (WHERE ss.package_id IS NOT NULL) AS selections
```
tarkah, line 115: I'll be curious what json aggregation performance is like vs separate queries -> rust side collection. Not as big of a deal for this state DB, but there might be material changes +/- where we do it on the meta db side.

tarkah (commented):
Nice work @livingsilver94! Overall this is looking really nice. I've left some feedback, mostly around regressions introduced from not deleting certain records on `add / batch_add` and no longer batching inserts where possible.

Also some open question around optimizing our get / list paths where we aggregate related records via json vs separate queries as done previously.

moss/src/db/layout/mod.rs
livingsilver94: Ah yes, `MAX_VARIABLE_NUMBER`. I've left the variable defined but it's currently unused, I wasn't sure it's worth using it because my gut feeling says we're not inserting so many entries.

I'll restore the batching logic.

moss/src/db/layout/mod.rs
livingsilver94: > Why do we need a process shared DB? 

I thought it's beneficial to have a read-only database that is populated only once, for perfomance. In fact, tests that need to modify a database will create their own.

moss/src/db/meta/mod.rs
livingsilver94: Your query is the exact reason why I decided to use multiple JOINs: it's complicated! :) My intention was to trade some performance (even though I haven't run any benchmark, so this is still to be proven) in favor of simplicity.

moss/src/db/layout/mod.rs
livingsilver94: I think we can? I can't see any blocker off hand. To be honest, I focused on optimizing selection queries, neglecting insertions/deletions. Good point, I'll try the rarray approach here as well.

moss/src/db/layout/mod.rs
livingsilver94: I think there's not limit except those imposed by system's `malloc()`?

No limit is explain here, but the example code shows that the `carray` is one variable bind: https://github.com/sqlite/sqlite/blob/master/src/carray.c

SQLite's `malloc()` limit is specified here, and it's ~2 GiB: https://github.com/sqlite/sqlite/blob/master/src/sqliteLimit.h#L43 but it shouldn't matter because the array is allocated by application's code.

moss/src/db/meta/mod.rs
tarkah: It's funny, I find the CTEs much easier to reason about. The multiplicative joins are not only going to balloon the pre grouped result set, I also find it much harder to reason about.

I don't think "simplicity" should be the driver for our SQL queries however, especially considering how big an impact a poorly optimized query has on performance.

I'll test both versions against my moss install DB today and see what the difference is for single & many packages. Especially on a package with LOTS of deps / providers == large multiplicative join.

moss/src/db/layout/mod.rs
livingsilver94: > or add an ON CONFLICT DO UPDATE SET "upsert clause"

unfortunately this is not possible because `package_id` is not `UNIQUE` per our current schema. We need to CRUD the old way.

```text
---- db::layout::test::db_adds_multiple_entries stdout ----
Error: Dbms(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint")))
```

moss/src/db/meta/mod.rs
livingsilver94: I genuinely thought auto-deletion is active by default like in other socket-based DBMSs. Thanks!

moss/src/db/meta/mod.rs
tarkah: @livingsilver94 we moved this conversation to zulip. I assume you want to keep your implementation?

If so, I think theres a "bug". We need all LEFT joins otherwise we will filter out packages that have no licenses / providers. I realize this is inferred to be correct due to us requiring a license and always injecting a package provider, however it violates the db schema and theres no guarantee this is upheld in the future. 

I'd also defensively add the FILTER to all json_group_array calls since the multi-join will cause nulls values to occur across each column I believe?

livingsilver94: I'll use your CTE approach if you feel more confident about using that one, no problem. I'm happy you found out the two solutions are practically equivalent, that's the little victory I like to bring home :)

livingsilver94: I think the CTE-based query is harder to compose when filters must be applied. Your example filters by package name, but there are [other criteria](https://github.com/AerynOS/os-tools/blob/cb294081df02702efc76ea6752178acc3333b09b/moss/src/db/meta/mod.rs#L75) as well. Such filters must go inside the right subtable, whereas with my naive approach I just need to append WHERE and HAVING. Am I misunderstanding?

tarkah: Nope you're spot on. As I mentioned in Zulip I'm fine going ahead with your approach after verifying it's performance. I just had those couple last items I pointed out but otherwise I think it's fine.

livingsilver94: Done. Sorry for the delay.

tarkah (approved):
Pushed a few commits to cleanup a few things & fix a few regressions compared to the diesel version. Only one regression was a major behavior change 88b74140. 

Looks great, thanks for this!

PR #789, opened 2026-05-19
Fix issue where `moss sync` with `--import` does not use the repositories in the `system-model.kdl`

### LINKED ISSUE #779: Improve the system-model --import process in moss
## The Problem

A user pointed out that the current behaviour of --import differs from actually activating a system-model.kdl:

/etc/moss/system-model.kdl exists:

**Prerequisites:** _No repo added to moss._

```
sudo cp model.kdl /etc/moss/system-model.kdl
sudo moss sync -u
```
The repo in system-model.kdl will work.

But:

```
sudo moss sync -u --import model.kdl
Error: sync: sync: Package defined in system-model does not exist in any repository: name(pkgset-aeryn-base)
```
Here the import does not work.

## A Solution

I think that, when using `moss sync -u --import <some system-model.kdl>`, moss could:

1) Back up the state of existing repos
2) Disable all existing repos (no existing repos? no problem)
3) Add repos from the supplied model with the specified priorities. If any of the model specified repos fail to resolve, the model import should fail with an informative message. If repos already exist, they will be updated per the model.
4) Enable the repos from the system model (with the priorities from the model added above)
5) Do the actual sync. If any of the packages fail to resolve, the import should fail with an informative message, and the previous state of the existing repos should be restored.

We should also make sure to enable `moss sync -u --import <model> -D <destination root>` to work into both empty and non-empty, but moss should prompt first (with --yes being an override) and display useful information about the destination, such as "does it already have an /etc/os-release file, does it have a /.moss/ dir" etc.

lumi-me-not:
This also adds `ClientBuilder` and modifies `Client::apply_stateful_blit` to make sure `/etc` exists before running triggers.

Tested locally and everything works as expected. `cargo test --all` also succeeds.

Fixes https://github.com/AerynOS/os-tools/issues/779

### DIFF
--- a/boulder/src/build.rs
+++ b/boulder/src/build.rs
@@ -163,14 +163,10 @@ impl Builder {
 
         // Prune moss cache, retaining stones from the repos defined
         // by our boulder profile
-        {
-            moss::Client::with_explicit_repositories(
-                "boulder",
-                moss::Installation::open(&self.env.moss_dir, None)?,
-                self.repos.clone(),
-            )?
+        moss::ClientBuilder::new("boulder", moss::Installation::open(&self.env.moss_dir, None)?)
+            .repositories(self.repos.clone())
+            .build()?
             .prune_cache()?;
-        }
 
         Ok(())
     }
--- a/boulder/src/build.rs
+++ b/boulder/src/build.rs
@@ -163,7 +163,7 @@ impl Builder {
 
         // Prune moss cache, retaining stones from the repos defined
         // by our boulder profile
-        moss::ClientBuilder::new("boulder", moss::Installation::open(&self.env.moss_dir, None)?)
+        moss::Client::builder("boulder", moss::Installation::open(&self.env.moss_dir, None)?)
             .repositories(self.repos.clone())
             .build()?
             .prune_cache()?;
--- a/moss/src/cli/sync.rs
+++ b/moss/src/cli/sync.rs
@@ -56,12 +56,12 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
         client_builder = client_builder.system_model_path(path);
     }
 
-    let mut client = client_builder.build()?;
-
     // Make ephemeral if a blit target was provided
     if let Some(blit_target) = command.blit_target {
-        client = client.ephemeral(blit_target)?;
+        client_builder = client_builder.ephemeral(blit_target);
     }
+
+    let mut client = client_builder.build()?;
 
     // Update repos if requested
     if update {
--- a/moss/src/cli/sync.rs
+++ b/moss/src/cli/sync.rs
@@ -4,7 +4,7 @@
 use std::path::PathBuf;
 
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
-use moss::{ClientBuilder, Installation, client::Client, environment, runtime};
+use moss::{Installation, client::Client, environment, runtime};
 use tracing::instrument;
 
 pub use moss::client::Error;
@@ -50,13 +50,13 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
     let simulate = command.dry_run;
     let update = command.update;
 
-    let mut client = if let Some(path) = &command.import {
-        ClientBuilder::new(environment::NAME, installation)
-            .system_model_path(path)
-            .build()?
-    } else {
-        Client::new(environment::NAME, installation)?
-    };
+    let mut client_builder = Client::builder(environment::NAME, installation);
+
+    if let Some(path) = &command.import {
+        client_builder = client_builder.system_model_path(path);
+    }
+
+    let mut client = client_builder.build()?;
 
     // Make ephemeral if a blit target was provided
     if let Some(blit_target) = command.blit_target {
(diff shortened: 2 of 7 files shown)


boulder/src/build.rs
```diff
@@ -164,12 +164,10 @@ impl Builder {
...
-                "boulder",
-                moss::Installation::open(&self.env.moss_dir, None)?,
-                self.repos.clone(),
-            )?
-            .prune_cache()?;
+            moss::ClientBuilder::new("boulder", moss::Installation::open(&self.env.moss_dir, None)?)
+                .repositories(self.repos.clone())
+                .build()?
+                .prune_cache()?;
         }
```
jplatte: I wonder why there is a block around this statement. I'm pretty sure it would not change anything at all to remove the `{` + `}`.

moss/src/client/mod.rs
```diff
@@ -89,25 +137,17 @@ pub struct Client {
...
-        client_name: impl ToString,
-        installation: Installation,
-        repositories: repository::Map,
-    ) -> Result<Client, Error> {
-        Self::build(client_name, installation, Some(repositories))
+        Self::build(client_name.to_string(), installation, None)
     }
 
     /// Build a functioning Client for the given [`Installation`] and repositories
     fn build(
```
jplatte: Kind of weird that `Client` has both a `::new` and a `::build`. Maybe the callers of `build` can be refactored to use `ClientBuilder`?

Also, if that works, I think it would make sense to have a `Client::builder` function that is exactly identical to `ClientBuilder::new`. It's a few characters shorter and a common naming scheme in Rust libraries. (also means fewer imports, though it looks like `moss::Client` is usually referred to by its fully qualified path anyways)

moss/src/client/mod.rs
```diff
@@ -137,17 +150,23 @@ pub struct Client {
...
+        Self::build(client_name.to_string(), installation, None, None)
     }
 
     /// Build a functioning Client for the given [`Installation`] and repositories
     fn build(
         client_name: String,
-        installation: Installation,
+        mut installation: Installation,
         repositories: Option<repository::Map>,
+        system_model_path: Option<PathBuf>,
```
jplatte: Oh, you're adding a new parameter here and the only callsite that has to change is `fn new` above? Then I'd say this PR should just inline `Client::build` into `ClientBuilder::build` and have `Client::new` call into that.

moss/src/cli/sync.rs
```diff
@@ -50,7 +50,13 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
     let update = command.update;
 
-    let mut client = Client::new(environment::NAME, installation)?;
+    let mut client = if let Some(path) = &command.import {
+        ClientBuilder::new(environment::NAME, installation)
+            .system_model_path(path)
+            .build()?
+    } else {
+        Client::new(environment::NAME, installation)?
+    };
```
jplatte:
```suggestion
    let mut client_builder = ClientBuilder::new(environment::NAME, installation);
    if let Some(path) = &command.import {
        client_builder = client_builder.system_model_path(path);
    }
    
    let mut client = client_builder.build()?;
```

jplatte (commented):
Can't really comment about the logic changes here, but left some comments about the code style :)

boulder/src/build.rs
lumi-me-not: I also do not think it would change anything, I just kept them because they were there before. I'll remove them.

moss/src/client/mod.rs
lumi-me-not: Yeah, `Client::new` is used in a couple places. I refactored the other callsites but thought it might look weird to have `ClientBuilder::new(/* ... */).build()?`.

I do like the suggestion of having `Client::builder()`.

moss/src/client/mod.rs
lumi-me-not: Ok, will do.

moss/src/cli/sync.rs
lumi-me-not: Will do.

tarkah (approved):
Looks great! Builder pattern helped a lot. Thanks!

moss/src/cli/sync.rs
```diff
@@ -50,7 +50,13 @@ pub fn handle(args: &ArgMatches, installation: Installation) -> Result<(), Error
...
+    let mut client_builder = Client::builder(environment::NAME, installation);
+
+    if let Some(path) = &command.import {
+        client_builder = client_builder.system_model_path(path);
+    }
+
+    let mut client = client_builder.build()?;
 
     // Make ephemeral if a blit target was provided
     if let Some(blit_target) = command.blit_target {
```
jplatte, line 60: In this `if let` block, we call `client.ephemeral`. Is that equivalent to calling `client_builder.ephemeral` and moving `let mut client = client_builder.build()?;` after this block? That seems like it would be prefereable (can probably remove the `mut` on `client` then?).

jplatte (approved):

moss/src/cli/sync.rs
lumi-me-not: Yeah, it is better to call it on the builder.

Can't remove `mut` from the client though, because `sync` and `refresh_repositories` require it.

PR #793, opened 2026-06-02
crates/version_parse: Fix parsing version when it isn't the last part

joebonrichie:
Previously, we assumed the last part of the url would contain the version, however, some urls have have some suffix, such as -src, -Source, so on.

Instead, iterate of the last parts and look for digits to grok the version.


ermo (approved):
Sure, why not.

PR #796, opened 2026-06-05
boulder/macros: Add Bazel macros

K1ngfish3r:
Options from https://bazel.build/docs/user-manual
User has to provide fetch/build target in recipe


ermo (approved):
Looks fairly straightforward.

PR #800, opened 2026-06-15
Boulder: Various improvements

ReillyBrogan:
Note that using a single codegen unit instead of the (default) 16 can make compiles painfully slow.


ermo (approved):
I concur with @joebonrichie's assessement that this is good to go.

PR #803, opened 2026-06-17
Use snafu over thiserror in boulder::package

jplatte:
Slooowly converting everything over. No cleanups this time, just straight-up conversion of thiserror to snafu. Doesn't mean that these error types are good as-is, I'm just not particularily 'inspired' right now – killing time bc. I can't sleep yet as I slept through half the day.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -304,6 +304,7 @@ dependencies = [
  "serde_yaml",
  "sha2",
  "similar",
+ "snafu",
  "stone",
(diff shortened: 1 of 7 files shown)


ermo (approved):
> Doesn't mean that these error types are good as-is, I'm just not particularily 'inspired' right now

That may be so, but this still looks like a genuine QoL improvement, even if it's "just" incremental. Thank You. 🙏

PR #809, opened 2026-06-19
Improve trigger logging

jplatte:
I don't know if the `delete` field in `Handler::Delete` actually represents paths. It doesn't seem to be used in practice (`execute_trigger_directly` just has a `todo!()` branch for it).


ermo (approved):
Looks very elegant.

PR #810, opened 2026-06-19
Use rest pattern instead of repeated wildcard pattern

jplatte:
Tiny code style improvement.


ermo (approved):
Nicer to read indeed.

PR #814, opened 2026-06-25
Rework script parser

lumi-me-not:
I've reworked the script parser and evaluator in `stone_recipe`, splitting  it off into a new crate, `stone_script`. This PR also integrates `stone_script` into `boulder` so it can use the new machinery.

This change also means `boulder recipe macros` will finally show all built-ins! :tada:

Also tested by compiling these packages: (and possibly others I am not thinking of right now)
- awww
- bluetui
- brotli
- bwrap
- curl
- flatpak
- gitui
- grim
- imv
- iwd
- kwin
- linux-stable
- lzip
- mesa
- nano
- podman
- python-cryptography
- python-magic
- python
- ruby
- sddm
- slurp
- spice-gtk
- starship
- systemd
- tzdata
- uutils-coreutils
- xz
- vim
- vulkan-headers
- vulkan-loader
- wayland-protocols
- wayland
- webkitgtk
- weston
- wine
- xdg-dbus-proxy
- zstd


ermo:
Could you perhaps be persuaded to attempt to rebase this into more "logical" groups of feature additions...?

lumi-me-not:
Yeah, I'm planning to squash the commit history into a few commits. Will take a look at that the next time I sit down to work on this. :)

lumi-me-not:
Are you happy with this, or do you prefer I split things up more?

ermo:
> Are you happy with this, or do you prefer I split things up more?

One for @tarkah (and *maybe*  @jplatte? Only if you have the mental bandwidth Jonas, otherwise ignore)

boulder/src/build.rs
```diff
@@ -104,14 +102,16 @@ impl Builder {
...
-                    .values()
-                    .flat_map(|script| script.dependencies.iter().map(String::as_str))
-            })
-        })
+    pub fn extra_deps(&self) -> Vec<String> {
+        let mut ctx = stone_script::ScriptContext::new();
+        for target in &self.targets {
+            for job in &target.jobs {
+                for script in job.phases.values() {
+                    ctx.eval(&script.env, &script.expr).unwrap();
```
jplatte: `unwrap` outside of a test 😬

boulder/src/build.rs
```diff
@@ -247,14 +254,19 @@ impl Builder {
...
-                            script::Command::Content(content) => {
+                            stone_script::Command::Output { output: content } => {
                                 // TODO: Proper temp file
                                 let script_path = "/tmp/script";
-                                fs::write(script_path, content).unwrap();
+
+                                {
+                                    use std::io::Write as _;
+                                    let mut file = fs::File::create(script_path)?;
+                                    writeln!(file, "{prefix}\n{content}\n")?;
```
jplatte: Why not import the `Write` trait at the module scope?

crates/stone_script/src/context.rs
```diff
@@ -0,0 +1,344 @@
...
+                                trace: unwind_stack(&mut stack),
+                            });
+                        }
+                        self.dependencies.extend(action.dependencies.iter().cloned());
+                        stack.push(Frame {
+                            expr: &action.value,
+                            progress: 0,
+                        });
+                    } else {
+                        return Err(Error::Eval {
```
jplatte: I'd use an early return for this instead, likely through `.ok_or_else(...)?` on `env.actions.get(name)`.

crates/stone_script/src/context.rs
```diff
@@ -0,0 +1,344 @@
...
+                            trace: unwind_stack(&mut stack),
+                        });
+                    } else if let Some(value) = env.builtins.get(name) {
+                        self.output.push_str(value);
+                    } else {
+                        return Err(Error::Eval {
+                            source: EvalError::UndefinedBuiltin { name: name.to_owned() },
+                            trace: unwind_stack(&mut stack),
+                        });
+                    }
```
jplatte: The order here seems slightly awkward. Instead of

- Error path 1
- Happy path
- Error path 2

I'd put both error paths first, which I'd probably write here as:

```suggestion
                    if !self.always_allow_builtins && stack.len() == 1 {
                        return Err(Error::Eval {
                            source: EvalError::BuiltinsNotAllowed,
                            trace: unwind_stack(&mut stack),
                        });
                    }

                    let value = env
                        .builtins
                        .get(name)
                        .ok_or_else(|| Error::Eval {
                            source: EvalError::UndefinedBuiltin { name: name.to_owned() },
                            trace: unwind_stack(&mut stack),
                        })
                    
                    self.output.push_str(value);
```

crates/stone_script/src/env.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// Add a definition to this [ScriptEnv]
+    ///
+    /// ```rust
+    /// # use stone_script::{ScriptEnv, Definition, Expr};
+    ///
+    /// let mut env = ScriptEnv::new();
+    ///
+    /// env.add_definition("rustflags", Definition {
+    ///   doc: None,
+    ///   value: Expr::parse("-C opt-level=3").unwrap(),
```
jplatte:
```suggestion
    ///     doc: None,
    ///     value: Expr::parse("-C opt-level=3").unwrap(),
```

crates/stone_script/src/expr.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub fn dump_to_string(&self) -> String {
+        let mut output = String::new();
+        self.dump(&mut output);
+        output
+    }
+
+    /// Concatenate two exprs, one after the other
+    pub fn concat(&self, other: &Expr) -> Expr {
+        Expr {
+            fragments: self.fragments.iter().chain(other.fragments.iter()).cloned().collect(),
```
jplatte: Slightly shorter version, if you want:

```suggestion
            fragments: std::iter::chain(&self.fragments, &other.fragments).cloned().collect(),
```

jplatte (commented):
The tests are poetry 🥇
Skimmed the code and left some random suggestions.

boulder/src/build.rs
lumi-me-not: Drats! My plans to crash `boulder` have been ruined! :D

boulder/src/build.rs
lumi-me-not: Yeah, I agree it would be best to just put it in the module scope.

crates/stone_script/src/context.rs
lumi-me-not: Good idea! :3

crates/stone_script/src/context.rs
lumi-me-not: `ok_or_else` saving the day yet again. x)

It is a very neat function, though!

boulder/src/build.rs
lumi-me-not: I'll need to have a deeper look here later, as just returning a `Result<_, Error>` eventually gets into an issue with an infinitely-sized type. (because of the error type being mutually recursive)

lumi-me-not: Ok, I've untangled it in a way that's satisfactory. (at least to me)

lumi-me-not:
To test this, I have built a bunch of packages. The only regression there is left, that I know of, is some networking issue with building `podman`.

The changes required to make the tuning logic work are a bit awkward, because it forced me to let builtins be `Expr`, (not doing so introduces a circular dependency in that `ScriptEnv` must have `ScriptEnv` to be created) but for now I think that's fine. This should be solved in a future PR by integrating the tuning parameters better into the macros.

If it is desirable for me to also bolt this onto this PR, I can do that, but I believe it would be better to make sure this PR is good and does not introduce regressions first.

Reworking the tuning flags may also be a good opportunity to onboard someone, and I'd be open to mentoring someone for that.

lumi-me-not:
Do tell me whenever I should squash the commits. :)

lumi-me-not:
Squashed!

boulder/src/cli/chroot.rs
```diff
@@ -51,7 +51,7 @@ pub fn handle(command: Command, env: Env) -> Result<(), Error> {
         )
         .map_err(Error::BuildScript)?
         .expect("script always available for prepare phase");
-    let profile = &build::format_profile(&script);
+    let profile = &build::format_profile(&env)?;
```
tarkah: Let's keep this named `script` / `script: ScriptBundle` makes sense. `env` shadows `env: Env` above and `Env` is a first class concept in `boulder` which makes this extra confusing.

boulder/src/build.rs
```diff
@@ -247,14 +264,18 @@ impl Builder {
                                 // Restore ourselves as fg term since bash steals it
                                 ::container::set_term_fg(pgid)?;
 
-                                if breakpoint.exit {
-                                    return Ok(());
+                                if exit {
+                                    return Err(Error::BreakpointExit);
                                 }
```
tarkah, line 260: Good catch, we should bail entirely vs continuing w/ packaging, etc.

boulder/src/build.rs
```diff
@@ -370,36 +391,41 @@ where
...
+    }
 
-    let definition_vars = script
-        .resolved_definitions
-        .iter()
-        .map(|(identifier, var)| format!("d_{identifier}=\"{var}\"; export d_{identifier}"))
-        .join("\n");
+    for (identifier, definition) in &script.env.definitions {
+        let mut ctx = stone_script::ScriptContext::new();
+        ctx.always_allow_builtins = true;
```
tarkah: This would be a nicer builder method... `ScriptContext::new().with_allowed_builtins()`

boulder/src/package.rs
```diff
@@ -119,10 +119,38 @@ fn resolve_packages(
...
+            doc: None,
+            value: stone_script::Expr::parse("%(_release)")?,
+        },
+    );
+
+    fn parse_content(env: &stone_script::ScriptEnv, source: &str) -> Result<String, Error> {
+        let mut ctx = stone_script::ScriptContext::new();
+        let expr = stone_script::Expr::parse(source)?;
+        ctx.eval(env, &expr)?;
+        Ok(ctx.flush_to_string())
```
tarkah: Might be nice to have a oneshot `Expr` method like `Expr::eval_to_string(self, env) -> String` that shortcuts this so caller doesn''t need to construct / discard the single use context

boulder/src/build/script.rs
```diff
@@ -0,0 +1,13 @@
...
+/// A bundle of a script and its execution environment
+#[derive(Debug)]
+pub struct ScriptBundle {
+    /// The environment to run the script in
+    pub env: stone_script::ScriptEnv,
+    /// Prefix to be added to the script before running it (previously called `env`)
+    pub prefix: stone_script::Expr,
+    /// The parsed script
+    pub expr: stone_script::Expr,
+}
```
tarkah, line 62: I think it'd be more ergonomic to pre-compute the prefix / commands within `ScriptBundle` for each phase, vs doing it at execution time. For one, we can cache things like dependencies which cause us to evaluate the script multiple times.

Secondly, it isolates all use of `stone_script` within a single place vs having to work w/ that API across multiple places.

We can then even move `format_profile` as a private function of that module / invoked by `ScriptBundle::new` and then we have a fully parsed script in one place that has:

- parsed script `.profile` string
- parsed script prefix
- parsed script commands
- parsed script dependencies 

And only a single place where evaluation occurs -> errors can occur

tarkah (commented):
This looks awesome, thanks @lumi-me-not!

I have reviewed the `boulder` changes and everything looks good, just a few comments. 

I still need to review the `stone_script` crate.

boulder/src/cli/recipe.rs
```diff
@@ -504,32 +504,47 @@ async fn fetch_and_cache_upstream(env: &Env, uri: Url, mpb: &MultiProgress) -> R
...
+                .unwrap_or_else(|| definition.value.dump_to_string())
+                .lines()
+                .next()
+                .unwrap_or_default()
+                .to_owned(),
+            example: None,
+        });
     }
     definitions.sort();
-    definitions.dedup();
```
tarkah, line 532: Unnecessary sort now that `env.definitions` are BTreeMap on String / name. We can also then remove Ord from PrintMacro which is awkward.

crates/stone_script/src/context.rs
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    fn finish(&mut self) {
+        if !self.output.is_empty() {
+            let command = Command::Output {
+                output: mem::take(&mut self.output),
+            };
+            self.commands.push(command);
+        }
+    }
```
tarkah: finish is a bit misleading here. Maybe `push_output` or more explicitly `push_output_command`?

boulder/src/build.rs
```diff
@@ -370,36 +391,41 @@ where
...
 
-pub fn format_profile(script: &Script) -> String {
-    let env = script
-        .env
-        .as_deref()
-        .unwrap_or_default()
+pub fn format_profile(script: &ScriptBundle) -> Result<String, Error> {
+    let mut prefix_ctx = stone_script::ScriptContext::new();
+    prefix_ctx.always_allow_builtins = true;
+    prefix_ctx.eval(&script.env, &script.prefix)?;
```
tarkah: Why would `prefix` have builtin's directly defined? From what I can tell, its recipe `environment` + `%scriptBase` and `%scriptBase` is an action so this flag wouldn't be needed here?

tarkah (commented):
Ok I just finished reviewing `stone_script`, beautiful! No major comments either. The design is really well thought out. I like the expression parsing into AST, explicit stack recursion / unwinding, and the split between parsed immutable env vs mutable context.

The new flow of env from macros -> parse expr -> eval w/ context flows nicely.

boulder/src/build.rs
lumi-me-not: I renamed `always_allow_builtins` to `allow_toplevel_builtins`, which is more specific, then I added `.with_toplevel_builtins()`.

tarkah (approved):
LGTM, thanks! Feel free to squash / amend however you like.

lumi-me-not:
Will do that when I am home again later today. :)

Let's Get That Moss!

lumi-me-not:
Squishy squashed! :3

tarkah:
> Squishy squashed! :3 

Please tell me you `jj absorb` (where possible) xD

lumi-me-not:
Not a `jj` user. xD One day I might try it out.

tarkah:
> Not a `jj` user. xD One day I might try it out.

It's a game changer imo, especially since you appear to prefer the interdiff / git ammend way of doing things. jj will make your life so much easier for wrangling your git history via conceptualizing things as "changes". Addtl bonus points if you find yourself working with lots of parallel or stacked branches and keeping them easily rebasable. Conflicts as a first class / deferrable concept is so good.

ermo:
@tarkah Is this ready for wider testing now in your view ...?

lumi-me-not:
> Is this ready for wider testing now...?

Yup!

Edit: Did not see the mention and thought it was directed at me. x3

tarkah:
Yup! I think it's safe to merge / test on volatile personally.

PR #822, opened 2026-07-09
Fix/moss hash checking

tarkah:
A smattering of (long overdue?) reliability and integrity improvements surrounding downloads and copies to the CAS, including:

- Fix state verify hash padding mismatch
- Unpack CAS files to .part w/ atomic rename
- Fix state prune partial path resolution
- Verify downloaded binary stone SHA256 against recorded upstream stone.index hash
- Verify hash when checking existing cached assets
- Verify CAS file hashes during unpack of downloaded .stones
- Verify CAS file hashes before skipping unpack when assets already exists in cache

### DIFF
--- a/moss/src/client/cache.rs
+++ b/moss/src/client/cache.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use snafu::{OptionExt, ResultExt as _, Snafu, ensure};
-use stone::{StoneDecodedPayload, StonePayloadIndexRecord, StoneReadError};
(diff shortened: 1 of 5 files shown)


ReillyBrogan (approved):

ermo (approved):
I suspect that the last commit in particular might fix a longstanding annoyance when you run `moss sync -u` with a newly rebuilt .stone in the local repo, which already exists in the same revision in the cache, and where the release number has not been updated, because this now force checks the hashes, not just if that release exists in the cache.

Previously, it was necessary to do an extra manual `moss repo update` first for this to work as expected.

PR #833, opened 2026-07-29
Format states as EROFS meta-only images

tarkah:
Allows us to format states as EROFS meta-only images, which can be `overlay` mounted w/ our moss CAS backing the image to provide deduped, immutable files w/ unique metadata.

This will eventually become a system-wide fstx driver, but for now this exposes it as `moss state format-erofs` as a means to test & dogfood.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,6 +1993,7 @@ dependencies = [
  "container",
  "dag",
  "derive_more",
+ "erofs",
  "fnmatch",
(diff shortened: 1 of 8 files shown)


ermo:
@tarkah looks like it needs a rebase after recent merges...?

tarkah:
> @tarkah looks like it needs a rebase after recent merges...?

On it

PR #834, opened 2026-07-30
moss: Upgrade native fstx driver

tarkah:
This commit applies a few major upgrades & fixes to our "native" fstx driver. I use "native" here instead of "hardlink" as we've previously described it since we now support "reflinks" where possible and also a fallback "copy" strategy when EXDEV / cross link.

The following notable changes are made:

- `reflinks` are now used when supported. This is an actual _correct_ implementation since we can apply `mode/uid/gid` bits on a per-metadata basis independently of the underlying shared CAS (content address store) blocks. This means `uid/gid` is applied for this strategy when blitting as root.
- `copy` is added as a fallback blit strategy only when EXDEV / cross link device error occurs (writing between separate filesystems). Previously this was an error since we could only `hardlink` and uses such as `moss it --to` would fail if writing out to some other filesystem. Like the `reflink` stategy, we can safely set `mode/uid/gid` bits here.
- The now `hardlink` strategy fixes a long standing safety issue (though no known package intersection is known to trigger this), where the `fchownat` call applying the incoming layout `mode` to the file can cause a file that already has its executability `x` bit set to lose its `x` bit. Since hardlinks all share an inode & the same metadata, we're really talking about `fchown` of the underlying CAS file, which is linked across ALL relevant states & possibly the running system root. A loss of `x` on a critical system binary linked into the root state can cause system unstability or crash. We now use a defensive strategy which ensures the `x` bit is always preserved (`old_mode | new_mode`) and only make the `mode` change when necessary.
- We now strip `w` bit on all modes applied from each blit strategy (`mode & 0o555`) since our `/usr` is intended to be read-only. We plan to further enforce this in the future via other means for true immutability.
- `umask` is always cleared prior to blitting so that `mode` bits are applied as-is for `openat` operations on the `reflink` and `copy` strategies, where `mode` can be defined atomically in `openat` when creating the destination file.
- Fixes a bug where `--to` had to exist ahead of time, even though we forcefully recreate the directory anyway.

### DIFF
--- a/moss/src/client/mod.rs
+++ b/moss/src/client/mod.rs
@@ -11,7 +11,10 @@
 use std::{
     borrow::Borrow,
     fmt, io,
-    os::{fd::RawFd, unix::fs::symlink},
+    os::{
+        fd::{AsRawFd, FromRawFd, RawFd},
+        unix::fs::symlink,
+    },
     path::{Path, PathBuf},
     time::{Duration, Instant},
 };
@@ -21,17 +24,18 @@ use fs_err as fs;
 use futures_util::{StreamExt, TryStreamExt, stream};
 use itertools::Itertools;
 use nix::{
+    NixPath,
     errno::Errno,
-    fcntl::{self, OFlag},
-    libc::{AT_FDCWD, RENAME_EXCHANGE, SYS_renameat2, syscall},
-    sys::stat::{Mode, fchmodat, mkdirat},
-    unistd::{close, linkat, mkdir, symlinkat},
+    fcntl::{self, AtFlags, OFlag},
+    libc::{AT_FDCWD, FICLONE, RENAME_EXCHANGE, SYS_renameat2, ioctl, syscall},
+    sys::stat::{Mode, fchmod, fstat, mkdirat, umask},
+    unistd::{Gid, Uid, close, fchown, mkdir, symlinkat},
 };
 use postblit::TriggerScope;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use stone::{StoneDecodedPayload, StonePayloadLayoutFile, StonePayloadLayoutRecord};
 use thiserror::Error;
-use tracing::{info, info_span, trace};
+use tracing::{info, info_span, trace, warn};
 use tui::{MultiProgress, ProgressBar, ProgressStyle, Styled};
 use vfs::tree::{BlitFile, Element, builder::TreeBuilder};
 
@@ -216,7 +220,7 @@ impl Client {
     pub fn ephemeral(self, blit_root: impl Into<PathBuf>) -> Result<Self, Error> {
         let blit_root = blit_root.into();
 
-        if blit_root.canonicalize()? == self.installation.root.canonicalize()? {
+        if blit_root.exists() && blit_root.canonicalize()? == self.installation.root.canonicalize()? {
             return Err(Error::EphemeralInstallationRoot);
         }
 
@@ -597,28 +601,9 @@ impl Client {
         }
 
         // Now swap staging with live
-        Self::atomic_swap(&usr_source, &usr_target)?;
+        atomic_swap(&usr_source, &usr_target)?;
 
         Ok(())
-    }
-
-    /// syscall based wrapper for renameat2 so we can support musl libc which
-    /// unfortunately does not expose the API.
-    /// largely modelled on existing renameat2 API in nix crae
-    fn atomic_swap<A: ?Sized + nix::NixPath, B: ?Sized + nix::NixPath>(old_path: &A, new_path: &B) -> nix::Result<()> {
-        let result = old_path.with_nix_path(|old| {
-            new_path.with_nix_path(|new| unsafe {
-                syscall(
-                    SYS_renameat2,
-                    AT_FDCWD,
-                    old.as_ptr(),
-                    AT_FDCWD,
-                    new.as_ptr(),
-                    RENAME_EXCHANGE,
-                )
-            })
-        })?? as i32;
-        Errno::result(result).map(drop)
     }
 
     /// Archive old states (currently not "activated") into their respective tree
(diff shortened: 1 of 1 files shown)


moss/src/client/mod.rs
```diff
@@ -1000,6 +985,138 @@ pub fn vfs(layouts: Vec<(package::Id, StonePayloadLayoutRecord)>) -> Result<vfs:
...
+    let from = installation.cache_path(".link-test");
+    let to = blit_target.join(".link-test");
+
+    let identify = || -> io::Result<_> {
+        // Remove both if they already exist
+        let _ = fs::remove_file(&from);
+        let _ = fs::remove_file(&to);
+
+        // Create empty from
+        let _ = fs::write(&from, b"");
```
jplatte: Do we really want to ignore (all) errors for these calls?

moss/src/client/mod.rs
```diff
@@ -1000,6 +985,138 @@ pub fn vfs(layouts: Vec<(package::Id, StonePayloadLayoutRecord)>) -> Result<vfs:
...
+    // Cleanup
+    let _ = fs::remove_file(&from);
+    let _ = fs::remove_file(&to);
+
+    strategy
+}
+
+/// syscall based wrapper for renameat2 so we can support musl libc which
+/// unfortunately does not expose the API.
+/// largely modelled on existing renameat2 API in nix crae
```
jplatte:
```suggestion
/// largely modelled on existing renameat2 API in nix crate
```

moss/src/client/mod.rs
```diff
@@ -1000,6 +985,138 @@ pub fn vfs(layouts: Vec<(package::Id, StonePayloadLayoutRecord)>) -> Result<vfs:
...
+fn atomic_swap<A: ?Sized + NixPath, B: ?Sized + NixPath>(old_path: &A, new_path: &B) -> nix::Result<()> {
```
jplatte, line 1061: Does `rustix` support this operation? Maybe not worth using if we don't already depend on it though (I don't know if we do). I've wanted to fully migrate to it before but had some trouble around the container code.

moss/src/client/mod.rs
```diff
@@ -1081,25 +1217,20 @@ pub fn blit_root(installation: &Installation, tree: &vfs::Tree<PendingFile>, bli
 /// Recursively write a directory, or a single flat inode, to the staging tree.
 /// Care is taken to retain the directory file descriptor to avoid costly path
 /// resolution at runtime.
-fn blit_element(
-    parent: RawFd,
-    cache: RawFd,
-    element: Element<'_, PendingFile>,
-    progress: &ProgressBar,
-) -> Result<BlitStats, Error> {
+fn blit_element(ctx: &BlitContext<'_>, parent: RawFd, element: Element<'_, PendingFile>) -> Result<BlitStats, Error> {
```
jplatte, line 1224: Might as well make this and other functions below methods on `BlitContext`? (maybe in a follow-up PR)

moss/src/client/mod.rs
```diff
@@ -1171,27 +1302,126 @@ fn blit_element_item(
...
+
+                            // Open FD to source
+                            let src = fcntl::openat(ctx.cache_fd, &fp, OFlag::O_RDONLY, Mode::empty())?;
+                            // Create new inode & get its FD.
+                            let dest = fcntl::openat(
+                                parent,
+                                subpath,
+                                OFlag::O_CREAT | OFlag::O_EXCL | OFlag::O_WRONLY,
+                                // We can specify mode here used to create the file.
+                                // Strip write bit.
```
jplatte: Duplicate comment. I'd suggest putting

```rust
let mode = item.layout.mode & !0o222;
```

(with maybe a comment also explaining _why_ the write bit is stripped?) above the `match *id` so it doesn't have to be repeated in both match arms.

Also maybe that empty-file hash should be a named constant?

moss/src/client/mod.rs
```diff
@@ -1171,27 +1302,126 @@ fn blit_element_item(
...
                 _ => {
-                    linkat(
-                        Some(cache),
-                        fp.to_str().unwrap(),
-                        Some(parent),
-                        subpath,
-                        nix::unistd::LinkatFlags::NoSymlinkFollow,
-                    )?;
+                    match ctx.blit_strategy {
+                        BlitStrategy::Reflink => {
```
jplatte, line 1327: Maybe each of these match arms could be its own function? The indentation level here is quite high.

jplatte (approved):

moss/src/client/mod.rs
```diff
@@ -1000,6 +985,138 @@ pub fn vfs(layouts: Vec<(package::Id, StonePayloadLayoutRecord)>) -> Result<vfs:
...
+                RENAME_EXCHANGE,
+            )
+        })
+    })?? as i32;
+    Errno::result(result).map(drop)
+}
+
+/// [`nix::unistd::linkat`] doesn't supply AT_EMPTY_PATH
+/// to allow `olddirfd` to be the source FD, so we override
+/// their implementation to support this usecase
```
joebonrichie, line 1079: Possibly fixed in newer nix crate versions, see
https://github.com/AerynOS/os-tools/pull/718/changes#diff-52e3d57a550ed261e78f12942e651576ee7279264de6bc37d62ced2a9b0a6294R164

tarkah: Agreed, though upgrading has some further reaching impacts than should be bundled on this branch.

I think we should do an honest effort to fully migrate to `rustix` across the board which will cover this.

moss/src/client/mod.rs
tarkah: Ahh yeah, I originally had these when I has the `identify().unwrap_or` falling back to `copy`, but that didn't make sense as it could _regress_our behavior to a _worse_ option than the existing hardlink functionality.

Now that I have any error falling back to hardlink to ensure 0 regression in behavior & only _improvement_, I think its safe to just bail on these errors within `identify` closure now.

moss/src/client/mod.rs
tarkah: Yeah I agree. We need a push to fully migrate to `rustix` and drop `nix`  & `nc`. We don't pull in `rustix` here yet so I think we can just postpone until we do that blanket refactor.

moss/src/client/mod.rs
tarkah: I was thinking of moving 100% of `blit` stuff to its own module / a bigger refactor. I think this will happen naturally as a consequence of the EROFS work as we are moving to a "driver" API where we have our existing "native" driver and the "erofs" driver. So I'm thinking we just hold off for that.

I wanted to keep this diff as simple as possible to ensure we catch any possible regression easily in review.

moss/src/client/mod.rs
tarkah: Same comment as https://github.com/AerynOS/os-tools/pull/834#discussion_r3691501457

ermo (approved):
I used this moss to successfully upgrade an old install from January 2025.

ermo:
@tarkah any chance of a rebase?

PR #837, opened 2026-08-01
Accept Git repository URLs in `boulder recipe new`

### LINKED ISSUE #537: Boulder: Add `git|` support in URIs supplied to `recipe new` command
> **Anon Ally**
>
> Hey there!
>
> For git based URL, is there a way to use boulder recipe new command ? The goal is to add mullvad vpn as package
> On solus packages repo, mullvad vpn use `git|repo` url
> **ermo**
> I don't think we currently support that.
> We should though.
> **ermo**
>
> the workaround is to use a tarball first, and then switch to the following format:

    upstreams   :
        - git|https://<git forge>/someproject/thecodebase.git :
            ref : <some long git ref>

livingsilver94:
This PR makes the `boulder recipe new` command accept Git-styled URI to create a recipe out of a Git repository. a Git-styled URI is a regular URL prefixed with our `git|` specifier. The recipe created points to the hash of the current HEAD of the repo.

I've also simplified URL parsing in `draft::monitoring::github` and `draft::monitoring::gitlab`, as in my opinion the regexes were hard to grasp. We're still leveraging regexes (although ideally we would ditch them), but they're shorter and only match the path inside the URL. Other checks are performed using functions provided by the `Url` struct.\
No functional change is introduced in these two modules *except* in gitlab's regular expressions, because the original ones weren't matching projects with a dot (".") in their name. In fact, I've also created a test with [mesa3d.org's URL](https://gitlab.freedesktop.org/mesa/mesa3d.org).

There's one effort remaining that is not covered in this PR: when running `boulder recipe new <first_url> <second_url>`, the command eventually fails because under the hood Git is trying to clone a repo inside the extraction root, that is not empty. Git is not happy about this. #838 will cover this case.

Fixes #537.

### DIFF
--- a/boulder/src/draft/upstream.rs
+++ b/boulder/src/draft/upstream.rs
@@ -19,7 +19,7 @@ use crate::Env;
 use crate::upstream::{self, git, plain};
 
 pub struct Upstream {
-    pub uri: Url,
+    pub uri: SourceUri,
     pub hash: String,
 }
 
@@ -34,16 +34,16 @@ pub fn fetch_and_extract(env: &Env, upstreams: &[SourceUri], extract_root: &Path
                 pb.enable_steady_tick(Duration::from_millis(150));
                 let upstream_dir = env.cache_dir.join("upstreams");
 
-                let upstream = match uri.kind {
+                let upstream = match &uri.kind {
                     stone_recipe::upstream::Kind::Archive => {
-                        fetch_and_extract_archive(&uri.url, &upstream_dir, extract_root, &pb).await?
+                        fetch_and_extract_archive(&uri.clone(), &upstream_dir, extract_root, &pb).await?
                     }
                     stone_recipe::upstream::Kind::Git => {
-                        fetch_git_repo(&uri.url, &upstream_dir, extract_root, &pb).await?
+                        fetch_git_repo(&uri.clone(), &upstream_dir, extract_root, &pb).await?
                     }
                 };
 
-                pb.suspend(|| println!("{} {}", "Fetched".green(), *uri));
+                pb.suspend(|| println!("{} {}", "Fetched".green(), uri.url));
 
                 Ok(upstream)
             })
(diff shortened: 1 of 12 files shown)


ermo:
Looks sane to me, but will let a Rust expert chip in.

boulder/src/draft/upstream.rs
```diff
@@ -1,74 +1,51 @@
...
-                let pb = mpb.add(
-                    ProgressBar::new_spinner()
-                        .with_style(
-                            ProgressStyle::with_template(" {spinner} {wide_msg}")
-                                .unwrap()
-                                .tick_chars("--=≡■≡=--"),
-                        )
-                        .with_message(format!("{} {}", "Downloading".blue(), *uri)),
-                );
+                let pb = mpb.add(ProgressBar::new_spinner());
```
tarkah, line 33: I think we dropped the Downloading message? I don't see it added downstream of here.

boulder/src/draft/upstream.rs
```diff
@@ -79,23 +56,6 @@ pub fn fetch_and_extract(env: &Env, upstreams: &[Url], extract_root: &Path) -> R
...
-        .map_err(Error::Bsdtar)?;
-    if result.status.success() {
-        Ok(())
-    } else {
-        eprintln!("Command exited with: {}", String::from_utf8_lossy(&result.stderr));
-        Err(Error::Extract(result.status))
-    }
-}
-
 pub fn fetched_upstream_cache_path(env: &Env, uri: &Url, hash: &str) -> PathBuf {
```
tarkah, line 59: Can we remove this? It'd be nice to use the upstream::plain module like you did below for this. I'm on mobile so I can't see if anything else references this.

tarkah (commented):
LGTM! Some nice cleanup in here as well. I have just two small comments.

boulder/src/draft/upstream.rs
livingsilver94: I let `upstream::plain` and `upstream::git` manage the progress bar, so yeah I should re-add the message there.

boulder/src/draft/upstream.rs
livingsilver94: No, it's used here: https://github.com/AerynOS/os-tools/blob/db9f41f189bfd63e967e92daac89f5e98f3cae13/boulder/src/cli/recipe.rs#L485
I did not touch `boulder recipe update` because I think it's out of the scope.

You're reviewing PR from a phone? Legend.

tarkah: Ahh ok, all good then.

Haha yeah, my cat trapped me on my bed 😆

tarkah (approved):

ermo:
@livingsilver94 did you run `just test` with the MSRV compiler version?

Because I get failures w/rustc 1.91.1 that look like this:

image

livingsilver94:
No, I tested it with `rustc 1.97.1 (8bab26f4f 2026-07-14) (Fedora 1.97.1-1.fc44)`. That is a crate I did not touch though. Want me to fix the error anyway?

ermo:
> No, I tested it with `rustc 1.97.1 (8bab26f4f 2026-07-14) (Fedora 1.97.1-1.fc44)`. That is a crate I did not touch though. Want me to fix the error anyway?

If possible, that would be nice. =)

PR #838, opened 2026-08-01
Ask for confirmation when `boulder recipe new` is passed multiple upstreams

livingsilver94:
This PR adds some more documentation and functionalities to `crate::upstream` and removes the `Upstream` struct in `draft::upstream`.

It also makes `boulder recipe new` ask for confirmation when multiple upstreams are passed, since only the first one will be scanned for the build system, licenses and dependencies.

Depends on #837, please review and merge that one first.

### DIFF
--- a/boulder/src/draft/upstream.rs
+++ b/boulder/src/draft/upstream.rs
@@ -2,29 +2,24 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use std::path::PathBuf;
-use std::{io, path::Path, process::ExitStatus, time::Duration};
+use std::{path::Path, time::Duration};
 
 use fs_err::tokio::{self as fs};
 use futures_util::{StreamExt, TryStreamExt, stream};
 use moss::{environment, runtime, util};
 use sha2::{Digest, Sha256};
+use stone_recipe::upstream::Kind as SourceKind;
 use stone_recipe::upstream::SourceUri;
 use tempfile::NamedTempFile;
-use thiserror::Error;
-use tokio::process::Command;
+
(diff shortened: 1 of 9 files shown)


ermo (commented):
Could you add an example screenshot/code text box showing the change in action?

livingsilver94:
[asciicast](https://asciinema.org/a/ekZW81wFkf6eDBJX)

livingsilver94:
Hm, it looks like `stone.yaml` gets recipes listed in alphabetical order, not in the order the user listed them. That's most likely a previously existing behavior.

boulder/src/draft/upstream.rs
```diff
@@ -56,7 +47,26 @@ pub fn fetch_and_extract(env: &Env, upstreams: &[SourceUri], extract_root: &Path
     ret
 }
 
+/// Extracts the upstream into the destination directory.
+/// The destination directory must exist.
+pub async fn extract(env: &Env, upstream: &Upstream, extract_root: &Path) -> Result<(), Error> {
```
tarkah: Super nitty, but fetch above encapsulates the runtime::block usage. Can we do the same here to drop the async fn so they're aligned?

tarkah (approved):
Lgtm! Nice split up of responsibilities and exposing composable primitives. 

One nitty comment. I haven't tested.

boulder/src/draft/upstream.rs
livingsilver94: It works!

ermo:
@tarkah Please merge when you're happy with this?

PR #839, opened 2026-08-02
Remove leftover Diesel references

livingsilver94:


ermo (approved):
Thanks for tying a bow on this, given you authored the rusqlite PR.

PR #840, opened 2026-08-02
Remove Serpent OS references

### LINKED ISSUE #494: Tidy up of AerynOS rebranding
I'm having a look around the os-tools repo and I'm seeing a lot of places where the header of a file has SerpentOS references rather than AerynOS. I've also seen this in a few readme's here on github (for example on the boulder repo readme. I'm not sure how easy it would be to capture this in one commit with some sort of find & replace function?

livingsilver94:
I replaced all Serpent OS references with AerynOS's.

Fixes #494.

### DIFF
--- a/funding.json
+++ b/funding.json
@@ -3,24 +3,24 @@
   "entity": {
     "type": "organisation",
     "role": "owner",
-    "name": "Serpent OS",
-    "email": "ikey@serpentos.com",
+    "name": "AerynOS",
+    "email": "ikey@aerynos.com",
     "phone": "",
-    "description": "The Serpent OS Linux distribution - a bold, independent project creating a new set of tooling to solve the problem of creating and managing distributions, effectively.",
+    "description": "The AerynOS Linux distribution - a bold, independent project creating a new set of tooling to solve the problem of creating and managing distributions, effectively.",
     "webpageUrl": {
-      "url": "https://github.com/serpent-os"
+      "url": "https://github.com/aerynos"
     }
   },
   "projects": [
     {
       "guid": "moss-and-boulder",
       "name": "tools",
-      "description": "Serpent OS package management tooling: moss and boulder",
+      "description": "AerynOS package management tooling: moss and boulder",
       "webpageUrl": {
-        "url": "https://github.com/serpent-os/tools"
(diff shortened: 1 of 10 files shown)


funding.json
```diff
@@ -3,24 +3,24 @@
   "entity": {
     "type": "organisation",
     "role": "owner",
-    "name": "Serpent OS",
-    "email": "ikey@serpentos.com",
+    "name": "AerynOS",
+    "email": "ikey@aerynos.com",
```
NomadicCore, line 7: @ermo should we change this to a different email address now? If so, I'm not sure which one would be best (considering scraping of data etc)?

funding.json
NomadicCore, line 41: I know this isn't part of this review, but we should also remove this now.

Not sure if it should be done in a separate PR though to keep things clean

funding.json
livingsilver94: I don't know, all I did is getting the email address from (dormant) Ikey's profile. Do we have an "institutional" email address now?

funding.json
livingsilver94: Honestly I think this PR is small enough to just add a new commit to it to remove this link, BUT, I think a relational change like that should not be buried under a PR that is not dedicated to it.

You're the PR person here, but I think it'd be better to remove that link "in broad daylight", for transparency. I think you get what I mean.

funding.json
NomadicCore: Yeah no worries, this was more pointed to Ermo, I should maybe have made that more clear.

We have a few different emails addresses but whichever we use and/or can make custom ones as needed.

This Ikey one would work, but it's more that with Ikey haven't stepped away, I don't know that we should keep using his name is all.

Keep it as is for now and lets wait for Ermo's guidance on how he wants to proceed.

funding.json
NomadicCore: Yeah, I think we should do it separately. I just saw it in looking at your PR.

We have already removed his sponsorship is most places back in October/November 2025 and also mentioned it in a blog post.

I agree though, separate PR to keep it clean and transparent.

ermo (approved):

ermo:
@livingsilver94 Thank You!

PR #842, opened 2026-08-04
crates/triggers: Support argument expansion with :all/:each

joebonrichie:
Currently, trigger handlers execute once per matched path, this can be slow for trigger handlerss which support processing multiple args.

Introduce argument expansion which allows triggers to batch arguments

- $(var) / $(var:each) - Execute handler per match
- $(var:all) - Execute handler against batched matches

e.g. instead of
systemctl preset foo.service
systemctl preset bar.service
systemctl preset baz.service

Do:
systemctl preset foo.service bar.service baz.service


joebonrichie:
Alternative to #701, we get practically the same performance uplift with the `systemctl preset` handler changed to use `:all`

PR #843, opened 2026-08-04
moss: Refactor "blit" to new fstree native module

tarkah:
This is a precursor for supporting multiple different fstree formats / drivers, such as the in research
overlayimg format.

This doesn't yet add any Driver trait / API, that'll come once we have the alternative implementation.

For now, this just refactors existing blit code out of `client` into `fstree::native`, there should be 0 change in the actual code.


ermo (approved):
Let's get this in.

PR #844, opened 2026-08-04
Task/more native blit updates

tarkah:
I found a regression on our hardlink strategy where I left an FD open per item blit :skull: 

I also added the ability to force a blit strategy via `MOSS_FSTREE_NATIVE_STRATEGY` var and per discussions w/ @**ermo | Rune Morling** , ensure we up the processes nofile soft limit to equal its hard limit. This will bump it from `1024` to `524288` on aeryn install.

I profiled the max open FDs on my live system blit (after fixing leaky FD) with all 3 strategies and got the following:

- `reflink` = 224396 entries blitted in 1.38s (163.2k / s) w/ max fds (164)
- `hardlink` = 224396 entries blitted in 0.95s (236.7k / s) w/ max fds (138)
- `copy` = 224398 entries blitted in 1.49s (150.7k / s) w/ max fds (146)

So only `164` FDs open at once on a fairly large install and I didn't notice big changes when adding / removing some large packages so I feel theres probably some natural backpressure keeping this in check

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ nix = { version = "0.27.1", features = [
     "process",
     "mount",
     "hostname",
+    "resource",
     "signal",
     "term",
 ] }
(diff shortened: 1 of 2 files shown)


ermo (approved):
Very nicely done, thank you!

PR #848, opened 2026-08-06
Check state asset hashes during verify

### LINKED ISSUE #847: moss: Adapt `moss state verify` to native reflink strategy
`moss state verify` doesn't work as expected for reflink based roots.

It was built w/ the assumption of hardlinks so it only content checks the CAS. But since the state reflink file can be mutated -> Cow -> different, we need to technically check both CAS and all files in each state independently of each other.

tarkah:
`state verify` only performed content hash checks on CAS assets, not assets installed into states.

Previously we only hardlinked so this somewhat made sense. Now with reflink or copy strategies, this is a larger gap since any mutation to a state asset wont bleed through to the CAS & get caught there.

Even w/ hardlink, someone could replace a file entirely which was still a gap.

So regardless of strategy, it's best for us to reverify all states at the expense of addtl. hashing.

I've also shipped a fix to make states returned from DB deterministically sorted by id. I noticed this when testing this branch, that each run would verify a different state first each time.

### DIFF
--- a/moss/src/client/verify.rs
+++ b/moss/src/client/verify.rs
@@ -4,15 +4,15 @@
 use std::{
     collections::{BTreeMap, BTreeSet},
     fmt, io,
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
 
 use astr::AStr;
 use fs_err as fs;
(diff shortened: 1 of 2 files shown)


ermo (approved):
LGTM, thanks!

PR #858, opened 2026-08-11
moss: Use moss assets dir for blit strategy identification

tarkah:
We have a regression in blit strategy caused by the new identification method when moss is invoked with `moss --cache` and the specified cache dir is located on a separate filesystem, it then falls back to copy.

Using cache dir as the _source_ of the identification check function is incorrect. It needs to use the same assets path that we link CAS files from so that it identifies the appropriate blit strategy.

We figured this out because our ISO generation process uses `moss --cache` on one FS and `-D` on `/tmp` which caused `copy` to get used and balloon that root.


ermo (approved):
🚀

PR #859, opened 2026-08-11
moss: Don't nuke blit_target, only nuke usr/

### LINKED ISSUE #857: Change how detached ("ephemeral") roots work
Currently, when `moss --directory $ROOT sync --to $TARGET` is run, in the blitting phase it will `rm -rf $TARGET` and then blit into `$TARGET/usr`. `.moss` is located in `$ROOT/.moss` in this case.

I suggest instead of `--to` and `--directory`, we use `--moss-dir` and slightly change how `--directory` works:

  - `--directory $ROOT` will be the target that is blitted to, where triggers run, etc
  - `--moss-dir $MOSS_DIR` will be the `.moss` directory

By default, `--directory $ROOT` will implicitly set `--moss-dir $ROOT/.moss`, so that behavior stays the same. If the moss directory is not on the same filesystem as the root, then this should emit a warning, as doing that is very inefficient.

What's nice with this terminology is that it mirrors `GIT_DIR` and `GIT_WORK_TREE`: `GIT_DIR` is the same as `--moss-dir` and `GIT_WORK_TREE` is the same as `--directory`.

I think it would be good to also rename ephemeral roots to detached roots, or something similar.

For `aerbox` I want to have the moss dir be in `$XDG_DATA_HOME/aerbox/moss`, and this is currently not possible, as it's not possible to set the location of `.moss` directly. Best I can do here is `$XDG_DATA_HOME/aerbox/moss/.moss`.

tarkah:
This doesn't matter when blitting to `staging` when moss is creating stateful fstrees that need to be activated. It impacts use of `moss it --to` when blitting to a detached root.

This allows `moss it --to` to update the detached root usr/ & run system triggers on that detached root without nuking all other persistent data on that detached root, such as if that root is used via some container.

CC: @lumi-me-not

### DIFF
--- a/moss/src/client/verify.rs
+++ b/moss/src/client/verify.rs
@@ -23,6 +23,7 @@ use crate::{
     package, runtime, signal, state,
 };
 
+#[allow(clippy::branches_sharing_code)]
 pub fn verify(client: &Client, yes: bool, verbose: bool) -> Result<(), client::Error> {
     println!("Verifying assets");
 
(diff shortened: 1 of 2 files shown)


ermo (approved):
Let's give it a go.

PR #860, opened 2026-08-12
moss: Handle unsupported repo versions

tarkah:
Unsupported repo versions were currently stubbed to a branch which intends to hook into a future `self-upgrade` flow that isn't defined yet. This means that all moss clients from v0 inception through now, if faced with a future format bump, will panic instead of seamlessly upgrading.

In-lieu of finising that self-upgrade flow, we can instead ensure moss instead sync's to the latest supported format version. This would allow moss to still sync to a version which includes the newer format moss, which would allow the user to sync again until they finally land on the appropriate version.

The only downside with this approach is it requires the user invokes `moss sync` as many times as needed until no more updates are available. But they at least have a path to do so.

The `self-upgrade` flow intends to automate this so moss will automatically re-invoke the newer moss until it lands on the desired version. That is a larger change and still pending development.

### DIFF
--- a/moss/src/repository/format/root_index.rs
+++ b/moss/src/repository/format/root_index.rs
@@ -28,6 +28,22 @@ impl RootIndex {
         self.get_history(ident).map(|meta| (ident, meta))
     }
 
+    /// Returns the defined `upgrade_via` history for the specified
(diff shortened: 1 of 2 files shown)


ermo (approved):
Good call.

PR #861, opened 2026-08-15
boulder/data: Modernize bolt macros

joebonrichie:
Update optimization flags:
- Replace deprecated hfsort+ with cdsort
- Replace deprecated -icf=1 with -icf usage
- Remove unneeded -use-gnu-stack option, recent binutils handle this fine https://forge.sourceware.org/binutils-gdb/binutils-gdb-mirror/commit/bbc969306f8d5fb6c7b636e25f6f8e278946ef23

Fix use of unbounded variables due to `set -u` being set, take an array instead.


ermo (approved):
Let's give it a go.

PR #874, opened 2026-09-13
boulder: Fix PGO builds not running when profile overrides are present

joebonrichie:
Check for the workload key against both the target build and the root build to ensure per-arch PGO builds are performed when profile overrides are present.


tarkah (approved):
Nice catch, thanks!

PR #876, opened 2026-09-22
boulder/data: Use PID to save profiled data when using clang PGO

joebonrichie:
This fixes occassional 'Unable to merge profile data file already exists' warnings when instrumenting with clang

Note, we use %%pid here to escape '%p' from boulder attempting to parse it


joebonrichie:
Appears to not be necessary for now, was sidestepping a previous issue where stale profiling data could affect pgo builds.

